PR #701, opened 2014-03-19
Performance

bstst:
Hi,

Firstly, resolved paths were not cached, so I added a fix for that (1ea7c1e). Although the performance win for this one is quite minuscule.

Secondly -- derequire. Now THAT's a serious bottleneck. I suggest to make it optional, which is what 37d434c implements. In my case the bundle performance went from 13-ish seconds to 2 seconds with derequire disabled. And I never had any issues in our huge code-base with derequire off.

Cheers,
M.


zertosh:
FWIW, derequire@2.0.0 is super fast now (see https://github.com/calvinmetcalf/derequire/pull/26)

PR #995, opened 2014-11-16
Ensure path to outfile exists before opening stream

jackwanders:
I found today that the `--outfile` option for browserify will fail when the user attempts to output the browserify output to a file in a non-existent directory. For example, when running 

``` bash
$ browserify ./my/src/script.js --outfile=./dist/bundle.js
```

the following error occurs when `./dist/` doesn't exist:

``` bash
events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: ENOENT, open './dist/bundle.js'
Warning: Command failed:
events.js:72
        throw er; // Unhandled 'error' event
              ^
Error: ENOENT, open './dist/bundle.js'
 Use --force to continue.
```

This seems like a common enough case (e.g. browserify may run during a build that contains a clean step that blows away the intended destination directory) that I believe browserify should simply make the directory rather than blow up when `fs.createWriteStream` fails.

This pull request uses `mkdirp` to ensure the outfile directory exists before opening the stream, and modifies a test case that would fail without this change.


rubennorte:
I'd also appreciate that this PR is merged. Now we have to manually create the output directories as a previous build step and is kind of annoying.

mysticatea:
:+1:

IanVS:
It's been more than a year, and this is still a problem.  Is there a reason this hasn't been reviewed or merged?

jcpst:
Would love this feature to be merged.

jasonkarns:
Hoping this gets updated and merged. This is still a problem. (Also worth noting that babel and rollup both happily create any necessary output directories, which means any pre-emptive directory creation in npm-scripts is solely to appease browserify.)

goto-bus-stop:
Oh yeah I think this is great — @jackwanders would you like to update the PR to current master? Otherwise I can do it too

jackwanders:
@goto-bus-stop I'd be happy to update this PR. Given the holidays, I trust waiting a few days is okay, given the age of the PR.

goto-bus-stop:
Yeah absolutely! Thanks!



On 28 December 2017 00:24:43 CET, Jack Anderson <notifications@github.com> wrote:
>@goto-bus-stop I'd be happy to update this PR. Given the holidays, I
>trust waiting a few days is okay, given the age of the PR.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/browserify/browserify/pull/995#issuecomment-354195173

-- 
Sent from mobile. Please excuse my brevity.

PR #1006, opened 2014-11-26
allow custom resolve function

tiii:
Allow a custom function to resolve require calls.
This functionality was present until version 4. Module-deps still supports it, but browserify does not pass the option.

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,6 +54,7 @@ function Browserify (files, opts) {
     self._pending = 0;
     self._entryOrder = 0;
     self._ticked = false;
+    self._delegateResolve = opts.resolve || bresolve;
 
     var ignoreTransform = [].concat(opts.ignoreTransform).filter(Boolean);
     self._filterTransform = function(tr) {
@@ -377,7 +378,7 @@ Browserify.prototype._createDeps = function (opts) {
     mopts.resolve = function (id, parent, cb) {
         if (self._ignore.indexOf(id) >= 0) return cb(null, paths.empty, {});
         
-        bresolve(id, parent, function (err, file, pkg) {
+        self._delegateResolve(id, parent, function (err, file, pkg) {
             if (file && self._ignore.indexOf(file) >= 0) {
                 return cb(null, paths.empty, {});
             }
--- a/readme.markdown
+++ b/readme.markdown
@@ -422,8 +422,9 @@ as the `opts.vars` parameter.
 `opts.externalRequireName` defaults to `'require'` in `expose` mode but you can
 use another name.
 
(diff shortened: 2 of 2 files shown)


jhansen-tt:
+1

jhansen-tt:
And please don't tell me that making an internal npm module is the solution.  That is NOT the solution.

This is the solution.

jmm:
@jhansen-tt 
> And please don't tell me that making an internal npm module is the solution.

I'm actually not sure what you're alluding to.

I personally am in favor of somehow exposing the resolver function so that a custom one can be supplied. I'm thinking about what's a good API for doing that. What I want is to be able to set it, but also access the default one so that I can wrap it. Full disclosure: I'm already setting a custom resolver in my [pathmodify](https://www.npmjs.com/package/pathmodify) plugin and want to legitimize it.

jhansen-tt:
Researching this topic (something like browserify + rootRequire) usually ends up where building internal npm modules is suggested.  See https://lostechies.com/derickbailey/2014/02/20/how-i-work-around-the-require-problem-in-nodejs/

Anyway, I think being able to hook require() in some way (even if it's only supported by gulp-browserify) would be extremely useful, and a proper way to resolve the '../../../../' require problems.

jmm:
@jhansen-tt Ok, now we're getting somewhere ;) I never heard of rootRequire before, I had to look it up. I think I have seen that article before. He may be a bit mixed up -- not sure, I just glanced at it for 2 seconds right now and saw the part that I think you're referring to.

So it sounds like your primary (or entire) interest in this is to solve the `../../../../` relative paths problem? One solution to that is to store or symlink your code in the `node_modules` directory. That has nothing to do with npm really. Addressing the relative paths problem is also the primary purpose of my [pathmodify](https://www.npmjs.com/package/pathmodify) plugin. As I mentioned, it currently overrides the default resolver function in an undocumented way, but I hope to legitimize that. If you don't like the `node_modules` solution, you may want to give pathmodify a shot and see if it works for you (if you have any questions or feedback, please share it in pathmodify's issue tracker). It provides some sugar over a custom resolver (and also allows controlling exposure).

See also http://stackoverflow.com/questions/22051663/get-browserify-require-paths-to-behave-more-like-requirejs/27609116#27609116.

Also, gulp-browserify is problematic. It's frowned upon by the gulp project, if that matters to you, and I believe it's unmaintained at this point.

jhansen-tt:
@jmm Yeah we actually aren't using gulp-browserify any more. Forgot about that.  Hopefully this feature can get merged, but I will also look at pathmodify.

Cactusbone:
+1

Issue #908, opened 2014-09-17
Can't expose require created with Readable stream


lukekarrys:
I'm trying to add a readable stream to a bundle and then require it by the exposed name in `main.js`.

**main.js**
```js
var bar = require('m1');
```

The following code produces an error `Error: Cannot find module 'm1'`.
```js
var Browserify = require('browserify');
var Readable = require('stream').Readable;

function stringStream(content) {
    var s = new Readable();
    s.push(content);
    s.push(null);
    return s;
}

var b = new Browserify();
b.require(stringStream('module.exports = 5'), {basedir: '.', expose: 'm1'});
b.add('./main.js');
b.bundle().pipe(process.stdout);
```

Should this work?

terinjokes:
@lukekarrys You've correctly exposed the module, and Browserify will use it when generating the bundle.

The error you're getting "Error: Cannot find module 'm1'" comes from `module-deps` which implements the node module resolution algorithm, and searches for "m1" in the node_modules directories, as well as by looking at the "browser" field in package.json. It doesn't know about any other modules you're dynamically generating and handing to Browserify.

What I recommend is to inform `module-deps` that it shouldn't attempt to resolve the module by using `b.exclude('m1');`. I've written up an example, with a few extra notes, [in a Gist](https://gist.github.com/terinjokes/9e95ea98d41919266fbb).

lukekarrys:
Thank you @terinjokes! That is soooo helpful :tada: 

I'm gonna go ahead and close this since I now get that the issue isn't in browserify.

terinjokes:
> That is soooo helpful :tada:

I try, and often fail, to provide helpful answers. I regret when I don't, but glad to have made you happy here. :smile:

PR #1030, opened 2014-12-17
Fix exposed requires being persisted across bundles

latentflip:
As mentioned in the comments on issue #908 https://github.com/substack/node-browserify/issues/908#issuecomment-55956552, `expose`d files were being persisted across calls to browserify. I tracked this down to a leak where modules were being appended to the (module) global builtins object. Fixed with a test, though I'm not sure if you want a test for such a fine grained bug, so feel free to remove it, but it does demonstrate the failure & subsequent pass with this fix.

### DIFF
--- a/index.js
+++ b/index.js
@@ -440,7 +440,7 @@ Browserify.prototype._createDeps = function (opts) {
     else if (opts.builtins && typeof opts.builtins === 'object') {
         mopts.modules = opts.builtins;
     }
-    else mopts.modules = builtins;
+    else mopts.modules = copy(builtins);
     
     Object.keys(builtins).forEach(function (key) {
         if (!has(mopts.modules, key)) self._exclude.push(key);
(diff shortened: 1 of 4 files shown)


terinjokes:
LGTM :+1: 

@substack ?

jmm:
I think the root cause of this might be a bug in module-deps, which I've been trying to get to the bottom of. I [asked about it](https://github.com/substack/module-deps/pull/65#issuecomment-70507321) in substack/module-deps/pull/65, but it was merged without comment (so far). Long story short, with the way I'm proposing resolving pathnames of exposed modules, it gets them treated as built-in (e.g. transforms aren't run). More info in substack/node-browserify/pull/1077. So I think something needs to change in module-deps re: that anyway, and it might solve the problem you're talking about as well.

substack:
I'll have a patch fix for this soon once I finish the v9 branch.

substack:
Published in 9.0.1. Thanks for the patch!

PR #1034, opened 2014-12-19
Allow requiring module locally and globally with `fullPaths=true`

### LINKED ISSUE #1020: -r fails when requiring the same file from multiple folders
Take this project for example:
/util.js 
/main.js --> require('./util'), require("./a/submodule")
/a/submodule.js --> require('../util')

Run: `browserify -r ./util.js main.js > bundle.js`

The bundle.js will not run because the resolved dependencies for util.js point to undefined. E.g.
`{"./a/submodule.js":1,"./util.js":undefined}`
`{"../util":undefined}`

Now run:`browserify -r ./util.js:./util main.js > bundle.js` 

The bundle.js still won't run but only one of the two resolved dependencies is undefined
`{"./a/submodule":1,"./util":"./util"}`
`{"../util":undefined}`

This seems like a bug to me, but I could just be misunderstanding this whole thing. A sample project to reproduce can be found here: https://github.com/marcboscher/browserify-test

Using browserify 7.0.0 on Windows 7

Thanks

gasi:
This fixes #1020 in conjunction with `watchify` which uses `fullPaths=true`.

**Update:** Turns out this breaks external global access to `require('a')`, e.g. an HTML page with `var a = require('a');` as `a` is now only indexed via `/absolute/path/to/multi_require/a.js`. Related to: https://github.com/substack/watchify/commit/3358625695ad17d4dc3ece639fd7b70c0693d743#commitcomment-9049286

Do you have any ideas on how to restore this behavior from Browserify 4.x/Watchify 0.10.x?

/cc @gscottolson @tarajane

### DIFF
--- a/index.js
+++ b/index.js
@@ -145,7 +145,11 @@ Browserify.prototype.require = function (file, opts) {
         )
     ;
     if (!row.id) {
-        row.id = expose || file;
+        if (!self._options.fullPaths) {
+            row.id = expose || file;
+        } else {
+            row.id = path.resolve(basedir, file)
+        }
     }
     if (expose || !row.entry) {
         this._expose[row.id] = path.resolve(basedir, file);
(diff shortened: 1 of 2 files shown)


nikku:
This has been resolved in browserify@9. You may close this PR.

PR #1043, opened 2014-12-26
Ensure transforms execute in added order.

terinjokes:
Resolves 6to5/6to5ify#17


blacksonic:
This fix breaks gulp builds https://github.com/substack/node-browserify/issues/1044

blacksonic:
Is ```process.nextTick``` important in the fix?

Issue #1038, opened 2014-12-22
How to NOT use the "through" module?


tustin2121:
Under the API section on using Transforms, the following statement is present:

> You don't need to necessarily use the through module, this is just a simple example.

And that's it. Nothing else mentions how one _could_ use a transform without using _yet another_ 3rd party module, it just says "you don't need to".

This is my major general issue with Browserify, actually: the fact that it can supposedly do all these wonderful advanced things, yet everything is so poorly documented, it's a hassle to try and figure out what you have to do to get things working. You go to the "help + articles" section of the site and there's 30 to 50 links to random articles instead of, say, actual documentation, enumerated about each feature, with at least an explanation as to what each option does if not examples. You know, something like the MDN. The handbook is vague at best, and I really do not want to click through every article in that list just to see if they cover some edge case I'm looking for.

As to the topic: No, I don't want to use "through" if I can help it. I just finished learning and creating Transform modules for node v10's streams, and it would be lovely if I could just use that. But everywhere in the "documentation", this "through" module is used, and looking at it, it uses a whole different API that's now out of date with node v10. And _I know_ dropping a node v10 Through class in there is simply going to break everything. So how do I not "need to necessarily use the through module"?

EDIT: More specifically, I'm writing this using the browserify API, as part of a node build script.js.

PR #1045, opened 2014-12-30
Make a small fix to documentation per #1038

tustin2121:
It would be nice if this could list more ways to use Transform streams, assuming there are more.


feross:
Thanks!

PR #1055, opened 2015-01-06
Issue #968 Fix

### LINKED ISSUE #968: Problem with standalone/require options
Hey! I have a little problem using the `standalone` option with `require`: having 2 modules (`a` and `b`), where `a` depends of `b`:

**b.js**

```
module.exports = {
    text: 'blabla'
};
```

**a.js**

```
var b = require('b');

module.exports = {
    init: function(){
        console.log(b.text);
    }
};
```

running `browserify a.js --standalone a --require ./b.js:b --outfile out.js` the result of `window.a` (created by the standalone option) is `Object { text="blabla"}` and not `Object { init=function()}`. 
Is this a known problem? any workaround?

Thanks!

SinisterMinister:
Fixes #968 by preventing code that's pulled in via `b.require()` to be considered the standalone module. 

**This breaks exposing a file added with `b.require` as the standalone module!**
### a.js

``` javascript
module.exports = "foo";
```
### Browserify Command

``` bash
$ browserify --standalone a --require ./a.js:a
```
### Browserify Output

``` javascript
!function(e){if("object"==typeof exports&&"undefined"!=typeof module)module.exports=e();else if("function"==typeof define&&define.amd)define([],e);else{var f;"undefined"!=typeof window?f=window:"undefined"!=typeof global?f=global:"undefined"!=typeof self&&(f=self),f.a=e()}}(function(){var define,module,exports;return (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({"a":[function(require,module,exports){
module.exports = "foo";
},{}]},{},[])(undefined)
});
```

When this code is run, it will throw a `MODULE_NOT_FOUND` exception as it tries to export the `undefined` module.

### DIFF
--- a/test/no_entry_require.js
+++ b/test/no_entry_require.js
@@ -1,0 +1,28 @@
+var test = require('tap').test;
+var browserify = require('../');
+var through = require('through2');
+var vm = require('vm');
+
+test('require with no entry file', function (t) {
+	t.plan(1);
+
+	var b = browserify();
+	b.require(__dirname + '/no_entry_require/a.js');
+	b.bundle(function (err, src) {
+		var c = { window: {} };
+		vm.runInNewContext(src, c);
+		t.equal(c.require(__dirname + '/no_entry_require/a.js'), "foo");
+	});
+});
+
+test('exposed require with no entry file', function (t) {
+	t.plan(1);
+
+	var b = browserify();
+	b.require(__dirname + '/no_entry_require/a.js', {expose: 'a'});
+	b.bundle(function (err, src) {
+		var c = { window: {} };
+		vm.runInNewContext(src, c);
+		t.equal(c.require('a'), "foo");
+	});
+});
--- a/test/no_entry_require/a.js
+++ b/test/no_entry_require/a.js
@@ -1,0 +1,1 @@
+module.exports = "foo";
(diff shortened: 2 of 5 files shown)


jmm:
@SinisterMinister I appreciate your effort on this. I'm not sure this is going to be the right fix though. For one thing, I have a [PR on browser-pack](https://github.com/substack/browser-pack/pull/51) to add the ability to create a standalone bundle and expose a `require()` at the same time, e.g.:

`bundler.js`
```js
browserify({standalone: 'whatever'})
  .require('entry', {expose: 'whatever'});
```

`index.html`
```html
<script src="bundle.js"><script>
<script>
// true
window.whatever === require('whatever');
</script>
```

But in any case, the example browserify command you gave should produce a working bundle shouldn't it? And your PR makes it produce a broken bundle?

pspeter3:
What's the status on this or the other fix?

PR #1057, opened 2015-01-07
Remove redundant conditional assignment.

jmm:
This conditional assignment appears to serve no purpose. Is it perhaps supposed to be `if (opts.transform === false) rec.transform = false;` like for the `! isStream(file)` code path later in the function? After either changing it to that or deleting it, all tests still pass and I'm not familiar enough with browserify's internals to know what role those properties are supposed to play (I just noticed the line while debugging something else).

### DIFF
--- a/index.js
+++ b/index.js
@@ -132,7 +132,6 @@ Browserify.prototype.require = function (file, opts) {
                 id: id
             };
             if (rec.entry) rec.order = order;
-            if (rec.transform === false) rec.transform = false;
             self.pipeline.write(rec);
             
             if (-- self._pending === 0) self.emit('_ready');


zertosh:
continuation of https://github.com/substack/node-browserify/pull/1190#issuecomment-96478435: I don't know if there was ever the ability to exclude a require from being transformed. But there is no such ability now in module-deps. In fact, the "transform" prop on a row is what module-deps uses to duck-type a transform vs a row.

I think that this and the other `if (opts.transform === false)` is safe to kill. `row` is always an `xtend` of `opts`, so the code isn't doing anything.

I've seen a few issues about ppl wanting to exclude a file from being transformed. IMHO that's a Bad Idea. That should be on the individual transforms to do - not browserify.

jmm:
@zertosh 
> I've seen a few issues about ppl wanting to exclude a file from being transformed. IMHO that's a Bad Idea. That should be on the individual transforms to do - not browserify.

I was thinking that might be useful. Let's say you have several transforms (babelify, etc.) you want to run on your application modules but that you have no need to run on some libraries or whatever that you want to bundle, that aren't in `node_modules`. The transforms would have no natural way to know you wouldn't want them running on those files, and even if they include a mechanism for white|black listing files that could get rather cumbersome to blacklist even one file for several transforms. Whereas if you could do something like: `browserify().require('whatever', {transforms: false})` you could handle it for that file for all transforms. What do you think?

zertosh:
Definitely useful. But how would you respond to someone that asks for away to only exclude or include certain transforms?

jmm:
@zertosh
> Definitely useful. But how would you respond to someone that asks for away to only exclude or include certain transforms? 

Good question, I was wondering about that. You could start contemplating more elaborate values for this hypothetical `transforms` prop, but needless to say that could be quite the rabbit hole. You could just say that if they don't want to exclude all transforms then they'll have the manage it themselves (in the way they'd have to anyway if this feature doesn't exist). That's what comes to mind right now.

zertosh:
I could be wrong, but aside from micro-optimizing your build, or dealing with a poorly written transform - is there a truly legitimate use? <-- this sounds terrible :(

zertosh:
Blah, I didn't mean to derail this PR. How about removing the other transform assignment so we can get this merged?

jmm:
> EDIT: Or are those legitimate enough uses that warrant expanding the API?

That's a good question. I don't know when it came along, but `noParse` was considered legit enough to include to optimize build time. There's no direct way to compare them unless you have a specific set of transforms to talk about, but I wonder how performance is affected by running transforms as opposed to parsing for deps / globals. If the difference is generally micro in comparison then it wouldn't make sense for performance reasons. It's also possible you wouldn't want a file transformed for other reasons, but I don't know if that use case is compelling enough to warrant it.

I'm still wondering what this line is about though. Did the feature of opting out of transforms already exist and this is a remnant? Or does it have something to do with transforms becoming pipeline rows? I was hoping someone who knew what it was about would chime in. Otherwise I think I should try to figure out what it's about before changing it.

> Blah, I didn't mean to derail this PR. How about removing the other transform assignment so we can get this merged?

You mean [this one](https://github.com/jmm/node-browserify/blob/redundant-line/index.js#L163)?

zertosh:
> You mean this one?

Yup that one.

> I'm still wondering what this line is about though

I couldn't tell you either. Even if it is a remnant of a possible earlier "exclude this file from transforms" option, the way mdeps works now, we'd have to pick another name for it anyway.

>  I wonder how performance is affected by running transforms as opposed to parsing for deps / globals.

Depends on the transform. Dep parsing happens in [node-detective](https://github.com/substack/node-detective/blob/00851b64d0a2b97891406aa306cdddd09c0d16f6/index.js#L72-L83). 99% of the runtime is spent on [building the ast](https://github.com/substack/node-detective/blob/00851b64d0a2b97891406aa306cdddd09c0d16f6/index.js#L6-L16). Traversing it once it's built takes a trivial amount of time. There's a [small optimization](https://github.com/substack/node-detective/blob/00851b64d0a2b97891406aa306cdddd09c0d16f6/index.js#L70) where if the source doesn't have the word "require", it won't even bother parsing. So if jQuery didn't include that word, it wouldn't get parsed. Sadly it includes the words "requires" and "required". I'm going to fix/improve node-detective to use a regex for `\brequire\s*\(` that can skip comment blocks (I need to think about it a bit more). After that change, everyone should see a speed boost.

@sebmck: What do you think about this:

One of the great inefficiencies of build systems today is that they pass raw source along between transformations, instead of an AST. If you could agree on an AST, and passed that along instead - wow, that would change the game. Agreeing on an AST and parser is a problem, but much less so after https://github.com/facebook/esprima/issues/111. We already use acorn, but a babel uses a fork of acorn. There is [talk](https://github.com/babel/babel/issues/858) of unforking/switching, but if we could partner up with babel and agree on one, and an API to pass it along, that would be so amazing. You build one AST, we use for dep parsing, babel uses it for whatever, and so do other transforms, and we'd have the fastest build pipeline ever.

sebmck:
@zertosh [ESTree](https://github.com/estree/estree) is already the pseudo-standard AST format.

zertosh:
@sebmck Do acorn and the acorn-babel fork generate the same AST? Could browserify generate it for its own dep parsing and then it off to babel?

sebmck:
@zertosh Well that's the goal of ESTree so yeah. Esprima, Acorn etc, all generate the same AST since they follow the ESTree spec. Babel can also directly consume an AST with `babel.transform.fromAST`.

zertosh:
@sebmck I totally missed `babel.transform.fromAST`. I'll run some experiments and report back :smile:

jmm:
@zertosh
> Even if it is a remnant of a possible earlier "exclude this file from transforms" option, the way mdeps works now, we'd have to pick another name for it anyway.

Because of duck-typing on `row.transform`? If a "file" row always has `id`and / or `file` and a transform row never does then you could overload the meaning of `transform`. Not necessarily saying *should*.

> Depends on the transform.

Yeah, it's like trying to compare apples to fruit salad.

> 99% of the runtime is spent on building the ast. Traversing it once it's built takes a trivial amount of time.

Thanks, that's good to know.

> There's a small optimization...I'm going to fix/improve...

Sounds like a good idea.

> You build one AST, we use for dep parsing, babel uses it for whatever, and so do other transforms, and we'd have the fastest build pipeline ever.

That sounds like a cool idea. It would require rewriting a lot of existing transforms right?

terinjokes:
> It would require rewriting a lot of existing transforms right?

And all of our dependencies. And probably a good amount of the plugins. It's my biggest concern.

substack:
Let's not derail this issue with passing an AST around (that idea has a LOT of problems).

jmm:
I researched this.

I think the line referenced by this PR was probably intended to be:

```js
if (opts.transform === false) rec.transform = false;
```

But I can't for the life of me figure out what these `row.transform|rec.transform|opts.transform` properties in `b.require()` were supposed to do. They were introduced in 78ca8b73edf9eda641e3a48d77d583c4ac16ffaa "fix order sensitivity of stream file test", landing in [5.7.0](https://github.com/substack/node-browserify/compare/5.6.1...5.7.0). They appear to be related to `b.ignore()` somehow, but I can't understand what they have to do with it or the referenced test, or the [changelog entry for 5.7.0](https://github.com/substack/node-browserify/blob/master/changelog.markdown#570). Transforms were not yet pipeline rows at that time.

<hr />

This illustrates that the way / degree to which the pipeline is documented makes it hard to tell what row properties are used for what purposes by what pipeline phases. There was a conversation here about whether this had something to do with module-deps, and that was a red herring, because this was before transforms were pipeline rows and before module-deps used `row.transform` for anything (as far as I can tell). But for all I know, any other pipeline phase could've used that property for something, or does now.

I'm working on pulling together documentation for the whole pipeline in one place so it's easy to get an overview of how rows flow through and what various phases do with the props. It's very much a work in progress, but here's what I have right now: http://jmm.github.io/browserify-pipeline-docs/

This needs to be documented better in any case, but I think this will help provide a foundationg for talking about ways that the pipeline mechanism can be simplified or improved. Trying to fill in this information highlights how convoluted some of the logic is, especially compared to what's documented about it. I hope that other people will want to help fill this in and if so I'll try to make that easy to do.

<hr />

Back to the subject of this PR, these `.transform` property assignments don't serve any purpose that I've been able to identify, at least contemporary to when they were added. Ironically it may now make sense to `delete row.transform` to prevent transforms from being added via `b.require()`. This currently adds a transform:

```js
var
  browserify = require('browserify'),
  row_flow = require('browserify-row-flow'),
  through2 = require('through2');

browserify('./entry')
  .plugin(row_flow().plugin())
  .require({transform: xform})
  .bundle()
;

function xform (file) {
  return through2.obj(function (chunk, enc, done) {
    this.push(Buffer(chunk.toString().toUpperCase()));
    done();
  });
}
```

zertosh:
> But I can't for the life of me figure out what these `row.transform|rec.transform|opts.transform` properties in `b.require()` were supposed to do.

Very puzzling. But thank you for the detective work.

> I'm working on pulling together documentation for the whole pipeline

Wow. Those docs are amazing! I wanted to add a few things but I couldn't figure what to edit to submit a PR :) Any chance it can be in markdown? Aside from documenting the current state, we should have place to put any proposed changes (like the `parseFilter` thing). There are def a few things I'd like to kill to simplify the API so we can build out cooler features (like your pathmodify should be part of core browserify). Anyway, please let me know how I can help with this. What parts can I document?

>  Ironically it may now make sense to delete row.transform to prevent transforms from being added via b.require()

I was actually going to suggest that we should pick off the properties that we need instead of just passing an entire copy along. Not only for code clarity, but to prevent unintended uses - undocumented features are not cool.

But yeah, remove the other `if (opts.transform === false) row.transform = false;` too so we can merge this already.

jmm:
@zertosh 

> Wow. Those docs are amazing! I wanted to add a few things

Thanks on both counts!

> but I couldn't figure what to edit to submit a PR :)

I didn't quite have that ready to roll yet. Here are the repos:
https://github.com/jmm/browserify-pipeline-docs-content
https://github.com/jmm/browserify-pipeline-docs-builder
https://github.com/jmm/browserify-pipeline-docs/tree/build

Following the readme in the last one will get the ball rolling.

> Any chance it can be in markdown? 

I don't think it would make sense to try to do the whole thing in markdown, but chunks of it could be if it made sense. I'm not sure that would provide much benefit for this though. Right now it's mostly little chunks of HTML in separate files. Take a look and let me know if you think there'd be some advantage to markdown (that outweights wrangling with it, of course).

> Aside from documenting the current state, we should have place to put any proposed changes (like the `parseFilter` thing).

I'm not too sure right now how to effectively do that in this format. What I had in mind is that there can be builds of this with a semver range attached so that you can query a browserify version number and get the build that documents that version. I figured hashing out potential changes would mostly happen in issues. But if there's an effective way to present that kind of information in this format there can obviously be builds for that.

> There are def a few things I'd like to kill to simplify the API so we can build out cooler features

That sounds good, although I'd say the thing I'm probably most interested in at this point is making everything more reliable and predictable. That of course would provide a more solid foundation for cool features.

> (like your pathmodify should be part of core browserify).

I appreciate the endorsement, but I have the feeling that would be a controversial proposal. I have had in mind to maybe eventually explore the idea of providing more first class support for that kind of thing, like maybe via a `required` event that's emitted at the beginning of module-deps' `walk()`, that would give a handler a crack at the `id` before resolution proceeds.

> Anyway, please let me know how I can help with this. What parts can I document?

Thanks! It would help anytime anyone can fill it in to whatever degree when the opportunity arises. There are props (by which I mean a combination of prop, phase, and input or output side with respect to the phase) that are incompletely documented, not described at all, or not even mentioned yet. The point of this is to be able to get an overview of how props flow through the pipeline all in one place to be able to identify issues like incongruities where phase A uses a certain prop in one way and phase D uses the same prop a different way, leading to problematic behavior. Like with respect to this PR, if you could scan this documentation and see that `b.require()` is doing something with `row.transform` and then module-deps is using that same prop to duck type transform rows, it could bring to light that something is out of whack. I definitely believe there are things out of whack with `id` and `file`, for example.

> I was actually going to suggest that we should pick off the properties that we need instead of just passing an entire copy along. Not only for code clarity, but to prevent unintended uses - undocumented features are not cool.

Hmm, I hadn't thought of that. I wonder if it could be useful to set arbitrary props for use in plugins / custom streams or for debugging though.

jmm:
@zertosh Superseded by #1262.

zertosh:
> Take a look and let me know if you think there'd be some advantage to markdown (that outweights wrangling with it, of course).

You get the builtin github rendering - like the README. Your docs look beautiful, but the amount of process involved could turn people away. I think the [gulp's docs](https://github.com/gulpjs/gulp/tree/master/docs) have worked out.

> What I had in mind is that there can be builds of this with a semver range attached so that you can query a browserify version number and get the build that documents that version.

Ah so you're think like a full on site for this. If it's just a series of markdown that lives in the repo, you get semver for free via tags :)

> That sounds good, although I'd say the thing I'm probably most interested in at this point is making everything more reliable and predictable

:+1: We can make things more reliable and predictable by killing things before we add stuff

> It would help anytime anyone can fill it in to whatever degree when the opportunity arises.

I have to figure out how to make changes first :stuck_out_tongue: 

> I wonder if it could be useful to set arbitrary props for use in plugins / custom streams or for debugging though.

For sure - it'll prob piss some people off, but the win of being able to understand what's going on, is totally worth it - just a thought.

jmm:
> You get the builtin github rendering - like the README.

Oh, I didn't realize you meant just check it in to a repo. You do get that, but you get no control over the styling, no scripting, no breaking it out into multiple files (and then combining them). I think to present this effectively needs a lot of HTML (and the ability to style it, and not HTML I'd want to have to write / maintain manually instead of using templating). So I think trying to do it all as a markdown document wouldn't be feasible.

> Your docs look beautiful, but the amount of process involved could turn people away. I think the gulp's docs have worked out.

I hear you. I don't think the process is all that bad though:

Once:
* Clone a repo
* Run a shell script

Repeat:
* Edit content files (HTML snippets organized in directories)
* Run a gulp task

I also think it's likely to be more advanced users who will even be interested in contributing to this. And, the content is in its own repo, so anyone can open an issue to say "The documentation for `b.require()` `inRow.id` should note that..." if they're not up to building it.

> Ah so you're think like a full on site for this.

Perhaps. I think I'd like to keep the scripting client-side if possible. But just to get the ball rolling with it I put it on github pages. That might be sufficient.

> If it's just a series of markdown that lives in the repo, you get semver for free via tags :)

That's true, but I'm not sure it'd be flexible enough. Could you update the documentation for an older version, or have a single copy of the document that covers version `x.x.x..z.z.z` and query for the documentation that applies to `y.y.y` (in the middle) without creating a tag for every release?

> I have to figure out how to make changes first 

Like I said, it shouldn't be too bad (and I'm open to streamlining it), but if you don't feel like waiting you can also open issues on the content repo.

> For sure - it'll prob piss some people off, but the win of being able to understand what's going on, is totally worth it - just a thought.

I had the thought that a compromise is another option. Declare a certain prop (let's call it `userData`) that you can put anything you want in, and all other props to be reserved and subject to any handling (e.g. stripping) that doesn't conflict with the documented API. That way you could provide a way to attach custom data and reduce issues stemming from undocumented use of props (e.g. by just whitelisting certain props like you said) and eliminate the potential for collisions. If this happened, then it would be wise to encourage using namespacing within that property. E.g. if I used it for my plugin, only mutate `userData.pathmodify`.

Issue #1039, opened 2014-12-23
Controlling exposure of bundled modules


jmm:
I'm trying to make sense of how to control exposure of modules in a bundle and finding the documentation and behavior difficult to follow.

Given browserify 7.0.3 or 8.0.3 and this directory structure:

```
- somedir
    bundle.js
    - src
        a.js
        b.js
        c.js
```
#1)

Is this situation supposed to break the bundle? (does for me):

``` js
// src/a.js
require('./b');
```

``` js
// bundle.js
browserify('./src/a')
  .require('./src/b')
```

Result (when bundle loaded in browser):

```
Uncaught Error: Cannot find module './b'
```
#2)

Is this situation supposed to happen?:

``` js
// src/b.js
module.exports = "This is b.js";
```

``` js
browserify('./src/a')
  .require('./src/c')
```

``` html
<script src="bundle.js"></script>

<script>
// Logs "This is b.js"
console.log(require(2));
</script>
```

jmm:
I've uncovered the source of problem #1. Details to follow.

jmm:
In case it's not clear, the details to follow were in #1059.

PR #1059, opened 2015-01-08
Properly resolve absolute path of exposed module.

jmm:
The change in #1033 is flawed. And the test coverage for requires is inadequate.

See #1039 (part 1). A broken bundle is generated when doing something like...

```js
// bundle.js
browserify('./src/a')
  .require('./src/b');

// src/a.js
require('./b');

// src/b.js
// I'm the file being required
```
...because the code in #1033 essentially does...

```js
this._expose[row.id] = '/somedir/src/b';
```
...which leads to a mismatch here...

https://github.com/substack/deps-sort/blob/c61e515015a0f8392c1b07530877084b6c411cb7/index.js#L72

...because there should be an entry `index['/somedir/src/b.js']`, but instead there's `index['/somedir/src/b']`.

I'm not 100% sure this pull request completely and correctly solves the problem for all cases, but it does solve it for the scenario in #1039 and is a step in the right direction. That this apparently used to work (according to @gasi in #1033) and that #1033 was merged as a fix underscores the need for better and more comprehensive tests around this functionality.

### DIFF
--- a/index.js
+++ b/index.js
@@ -151,7 +151,7 @@ Browserify.prototype.require = function (file, opts) {
         row.id = expose || file;
     }
     if (expose || !row.entry) {
-        this._expose[row.id] = path.resolve(basedir, file);
+        this._expose[row.id] = resolve.sync(file, {basedir: basedir});
     }
     if (opts.external) return this.external(file, opts);
     if (row.entry === undefined) row.entry = false;


jmm:
Here's an example of a test that currently fails:

[jmm/node-browserify@ae9aedb](https://github.com/jmm/node-browserify/commit/ae9aedba8b5dbcffc90aac2d49a1565b15160aae)

substack:
Merged in 8.1.1 with an asynchronous version. Thanks for the patch!

jmm:
@substack Thanks! I started working on an async version after I submitted this pull request and I think it was shaping up similar to what you did (although I've only glanced at that so far) -- I was mimicking the pattern used for adding transforms -- but then I realized I think there's an even better solution, since the paths are already being resolved later on. I'll flesh that out a bit more and submit a pull request so you can see what you think.

PR #1078, opened 2015-01-17
resolve exposed files with browser-resolve

### LINKED ISSUE #1072: 8.1.1 fails to resolve modules from "browser" field
~~Minimal reproduction: https://gist.github.com/omsmith/a6b8876a4354960a65f4~~

Minimal reproduction: https://github.com/omsmith/browserify-8.1.1-resolve-issue

Works fine in 8.1.0, but as of 8.1.1 `browser` entries don't appear to be picked up

omsmith:
fixes #1072 

Could previously `b.require()` names from the local package.json's _browser_ entry.

If this is desired behaviour (these entries can also be `b.external()`'d.), this restores it along with a simple test.

### DIFF
--- a/index.js
+++ b/index.js
@@ -152,7 +152,7 @@ Browserify.prototype.require = function (file, opts) {
     }
     if (expose || !row.entry) {
         self._pending ++;
-        resolve(file, { basedir: basedir }, function (err, res) {
+        bresolve(file, { basedir: basedir }, function (err, res) {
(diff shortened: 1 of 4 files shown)


jmm:
I don't know if this is actually broken, but the behavior you're asking for is restored by #1077. But if that behavior wasn't an accident in the first place, I'd say the tests (which you have) and documentation  need updating to reflect that.

PR #1082, opened 2015-01-21
Remove broken link to doc/changelog

nikolas:


feross:
thanks!

feross:
Test failures are not due to this PR which is documentation-only. The source of the problem is a change in `crypto-browserify`. I just sent a PR to fix it: https://github.com/crypto-browserify/randombytes/pull/1

PR #1096, opened 2015-01-26
ensure use of URL path separators

jden:
On Windows, URLs in sourcemaps are using the `path.sep === '\\'` character rather than the URL `/` character. This breaks folder support in Chrome dev tools, as in:

image

This PR ensure that URL separator characters are used instead when Browserify is run on Windows.


substack:
Thanks for the patch! Merged in 8.1.3.

PR #1100, opened 2015-01-29
Support requiring a shimmed module name from an external bundle

mwiencek:
If bundle `A` does  `require('shim')` (which is a reference to ./shim.js in the package.json's `browser` field), and bundle `B` does `external(A)` and tries to also `require('shim')`, then browserify currently throws a `MODULE_NOT_FOUND` error for the second require.

I'd expect that the short name would not only be available to `B`, but that it'd also be smart enough to see that it means the same external module in `A`. A workaround is using `expose` in conjunction with `external` directly on ./shim.js, but that's really unmanageable when you have tons of modules you want shared among multiple sets of bundles for different classes of pages.

I've added a patch and test case which solve this for me, but I don't know if it's robust or correct. If there's another way to achieve this I'd be grateful to hear it. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -200,9 +200,37 @@ Browserify.prototype.external = function (file, opts) {
     if (file && typeof file === 'object' && typeof file.bundle === 'function') {
         var b = file;
         self._pending ++;
+
+        var bdeps = {};
+        var blabels = {};
+
         b.on('label', function (prev, id) {
             self._external.push(id);
+
+            if (prev !== id) {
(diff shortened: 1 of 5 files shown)


mwiencek:
This currently doesn't work if you have another module `C` that does `external(B)`. At that point it seems to duplicate ./shim.js in C's bundle, and I haven't figure out how to solve that.

substack:
Thanks for the patch. Merged in 8.1.2.

PR #1105, opened 2015-02-04
Update "browser-pack" and remove unused "umd" dep

zertosh:
Update "browser-pack" https://github.com/substack/browser-pack/pull/49.

`umd` is actually used by `browser-pack`, and it has it listed as dependency already.


zertosh:
@substack Since this update to `browser-pack` has breaking changes for `browserify`, it'll likely merit a major version bump. If so, please remember to also bump `watchify`'s `browserify` dep :smile:

PR #1114, opened 2015-02-09
Update copyright year

lpww:
Updated copyright year for 2015. No other files were changed.
browserify copyright update


feross:
This license was copied from the original project - it doesn't make sense to update their license for them.

Issue #1124, opened 2015-02-15
config object transforms are run twice


simonzack:
Doing the following will run transforms twice, no matter what the transform is:

    var bundler = browserify('file.js', {
        transform: [to5ify]
    });

using `.transform` instead works.

seansfkelley:
Seeing this as well. Here's a more complete repro:

```javascript
var browserify = require('browserify');
var through = require('through');

function transform1() {
  console.log('this log line appears once');
  return through();
};

function transform2() {
  console.log('this log line appears twice');
  return through();
};

browserify('./blank.js', {
  transform: [transform2]
})
.transform(transform1)
.bundle();
```

substack:
All fixed in 9.0.2.

PR #1128, opened 2015-02-17
fix to #1124

simonzack:
`opts.transform` is already processed, copying it will result in it being processed twice.


substack:
Thanks for the patch, but this is tricky to review and might cause merge conflicts later because of all the whitespace changes. Can you send a patch without so many lines of diff?

simonzack:
@substack I've got rid of them now.

substack:
@simonzack Thanks a bunch!

PR #1139, opened 2015-02-24
Tests for problems introduced in ~resolve@1.1.5 and fixed in resolve@1.1.6

dominykas:
~~When a module is exposed, it should still resolve the way it would normally do, i.e. with/without extension and directories should fall back to index, and index from a directory should be accepted with/without extension too.~~

~~In this PR there are the tests that will fail - I am not able to run them reliably for some reason, so with `resolve@1.1.5` they simply hang, however they will pass if you `npm install resolve@1.1.2`~~

~~I'm not sure how to make them pass - that, I believe, belongs in the `resolve` module - I'll create a relevant issue there, mostly because I've no idea how things are wired up.~~

The following tests are added, where `x` is a file and `y` is a folder with an `index.js` inside:
- `b.require('./x', {expose: 'xyz'});`
- `b.require('./y', {expose: 'xyz'});`
- `b.require('./y/index', {expose: 'xyz'});`
- `b.require('./y/index.js', {expose: 'xyz'});`

### DIFF
--- a/test/resolve_exposed.js
+++ b/test/resolve_exposed.js
@@ -4,7 +4,7 @@ var test = require('tap').test;
 
 test('resolve exposed files', function (t) {
     t.plan(2);
-    
+
     var b = browserify(__dirname + '/resolve_exposed/main.js', {
         basedir: __dirname + '/resolve_exposed'
     });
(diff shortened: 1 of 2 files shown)


dominykas:
After rebasing tests now pass - I assume the tests are still worth merging into the suite?

ljharb:
@dominykas i know this is old, but if you could rebase it, I’d love to see it go in. It will help ensure that resolve doesn’t break behavior in browserify.

dominykas:
OK, I'll see what I can do.

dominykas:
Meh.

dominykas:
I've no recollection of details of what happened there, but let's see if rebasing makes the tests pass :)

PR #1175, opened 2015-03-20
Updating readme link to events module on npm

TehShrike:
The current Gozala module instead of the old shtylman module


stevemao:
:+1:

feross:
thanks, merged this in from another duplicate PR.

PR #1183, opened 2015-03-26
More OS-agnostic tests, less dust in temporary folder

ElNounch:
Use path.join() to get local styled filepath.
Use Temp package to create temporary folders on OS specific place,
plus wiping stuff after tests.
Provide node-executable location to spawn().


substack:
Thanks for the patches!

Issue #1044, opened 2014-12-30
Version 8.0.2 broke my build with gulp


burmisov:
Not sure weather this is a browserify issue, but versions prior to 8.0.2 used to work fine (including 8.0.1).

Now it exits with this error:
```
Error: write after end
    at writeAfterEnd (D:\code\webgis\node_modules\browserify\node_modules\labeled-stream-splicer\node_modules\stream-spl
icer\node_modules\readable-stream\lib\_stream_writable.js:161:12)
```

Here is my build code:
```js
var browserify = require('browserify');
var rename = require('gulp-rename');
var transform = require('vinyl-transform');
var reactify = require('reactify');
var size = require('gulp-size');
gulp.task('scripts-debug', function () {
    var browserified = transform(function (filename) {
        var b = browserify(filename, {
            debug: true,
            extensions: ['.js', '.jsx']
        });
        b.transform(reactify);
        return b.bundle();
    });

    gulp.src('src/js/app.js')
        .pipe(browserified)
        .pipe(rename('bundle.js'))
        .pipe(gulp.dest('build/js'))
        .pipe(size({ showFiles: true }));
});
```

cortesi:
Just a quick note that our build scripts have been broken by this issue too. So... bump.

romeovs:
Same here. works on `8.0.1` but not on `8.0.2`.

davidmerrique:
Same here :
```javascript
b.transform('coffeeify'); // Broken in 8.x.x
b.transform({ global: true }, coffeeify); // Works in 8.0.1, broken in .0,.2,.3
```

jmm:
I don't pretend to fully understand everything involved in this, but I believe what's happening is that as of browserify 8.0.2, when it finishes running transforms it causes the pipeline to `end()`, meaning that `bundle()` no longer returns a writeable stream, as vinyl-transform relies on. I will be curious to see what @substack has to say about this, but I believe there are at least 2 workarounds if you need to use browserify >= 8.0.2 in the meantime:

Where `entry`, `dest`, `b_opts`, and `some_xform` are whatever you want.

Refactor usage of vinyl-transform as described in hughsk/vinyl-transform#1:

```js
gulp.task('whatever', function () {
  var b, browserified;

  b = browserify(b_opts)
    .transform(some_xform);

  browserified = transform(function (filename) {
    b.add(filename);
    return b.bundle();
  });

  gulp.src(entry)
    .pipe(browserified)
    // ...
[...]

PR #1187, opened 2015-04-02
stream returned from b.bundle() should be readonly.

terinjokes:
Ensure the output of this method and event match the documentation, per #1044.


terinjokes:
@substack not sure why the "$NODE_PATHS" test fails in node 0.8.

ElNounch:
My bad for the failure in Node 0.8, not linked to your code at all, fixed in [https://github.com/substack/module-deps/pull/78].

substack:
merged in 9.0.5 and fixed the module-deps issue upstream

PR #1190, opened 2015-04-03
fix b.require(stream, opts)

zoubin:


jmm:
@zoubin Thanks for your submission. I asked about this same line in #1057. I never got around to tracing the logic to figure it out for myself. Can you please summarize what is broken before this and how this PR fixes it? Would you be interested in writing a test that would catch it if it breaks again?

zoubin:
@jmm I've checked the logic again, and found no good reasons to set `row.transform` in `.require`? According to the [api documentation](https://github.com/substack/node-browserify#brequirefile-opts), `.require(file)` is meant to make `file` available from outside the bundle. However, right now we can use it to add a transform:

```javascript
var b = require('./index.js')();
var through = require('through2');
var rs = new require('stream').Readable({ objectMode: true });
rs._read = function () {
    if (this._done) {
        this.push(null);
    } else {
        this.push('var a = "hello ";');
        this._done = true;
    }
};
b.require(rs);
b.require({
    global: true,
    transform: function (file) {
        return through.obj(function (row, enc, next) {
            row = new Buffer(row.toString() + '// comments added by transform');
            next(null, row);
        });
    }
});
b.bundle().pipe(process.stdout);
```

I prefer to restrict the properties of the object written into pipeline, like `.require(stream)`, but without adding the `.transform` property. Unless I misunderstand the purpose of `.require`.

zertosh:
wow. the fact that you can "add" a transform from `.require` is an implementation leakage. here is some browserify internals 101:

The transforms are applied in [module-deps](https://github.com/substack/module-deps), which also handles dep tracing and whatnot. A build is started by "seeding" the module-deps stream with the entry/required files and transforms. The things that are pushed into the module-deps stream are the "row" objects. module-deps does a bit of [duck-typing](https://github.com/substack/module-deps/blob/8e34e42433dfdf1389b9eb89086aa0604d8eca0a/index.js#L72-L91)[^1] on the rows to figure out if they're a transform or a file[^2]. 

So, the implementation leakage that you're seeing is from this transform-as-a-row biz that happens in [`Browserify.prototype.transform`](https://github.com/substack/node-browserify/blob/96d1662536cdf98f5c124c8d64616e77e3283c61/index.js#L271-L329). The code is a bit hard to read because it's got a few async steps happening - but the main points are:

  1. The transform row is [created](https://github.com/substack/node-browserify/blob/96d1662536cdf98f5c124c8d64616e77e3283c61/index.js#L308-L312).
  2. After we've figured out [where](https://github.com/substack/node-browserify/blob/96d1662536cdf98f5c124c8d64616e77e3283c61/index.js#L321-L325) the transform is (remember that you can pass a transform by name),
  3. They get [written](https://github.com/substack/node-browserify/blob/96d1662536cdf98f5c124c8d64616e77e3283c61/index.js#L291) to the "pipeline". At this point the pipeline is in ["recorder" mode](https://github.com/substack/node-browserify/blob/96d1662536cdf98f5c124c8d64616e77e3283c61/index.js#L554). All the "seed"[^3] rows are "recorded". They are re-played into module-deps when you call `bundle()`.

When you `b.require()` a file, it gets [written](https://github.com/substack/node-browserify/blob/96d1662536cdf98f5c124c8d64616e77e3283c61/index.js#L177) to the pipeline, just like a transform. And that's why @zoubin's example works.

@jmm This is something that we can fix with a well defined `row` structure.

[^1]: This `_transform` is a [stream `_transform`](https://nodejs.org/docs/v0.10.37/api/stream.html#stream_transform_transform_chunk_encoding_callback) function not a browserify source transformation function.
[^2]: At this stage, the rows are objects, but later on, dependencies are created as new rows but they're strings for a bit before they become row objects. The code calls them "files" so kinda confusing.
[^3]: That's not a thing. I made up that name. That's just how I reason about them.

jmm:
@zertosh 
> This is something that we can fix with a well defined row structure.

Yeah, that'll be nice. I've started work on some documentation to provide an overview of how things work currently to set the stage for discussing possible changes. The browserify internals 101 summary you gave is the kind of info I want to include in it.

It was before my time with browserify, but I think transforms weren't always rows. So perhaps that's how this situation arose. In the documentation I'm working on, in the "what is a row and how do you push one" section I was debating whether I should even mention transform rows or not (but I guess they need to be mentioned for developers inserting streams in the pipeline).

Thanks for the further info @zoubin. As @zertosh alluded to, we've been discussing bringing more documentation and consistency to the `row` props that control things like addition of files and transforms to the bundle, in #1203 for example. Because a lot of things aren't well documented right now it takes a considerable amount of digging through the code to try to understand how some things are intended to function and are actually functioning, and that's where I left off with this. I intend to do figure it out eventually and do what needs to be done, but @zertosh or someone else might before I do.

jmm:
@zoubin
> I've checked the logic again, and found no good reasons to set `row.transform` in `.require`. 
> ...
> I prefer to restrict the properties of the object written into pipeline, like .require(stream), but without adding the .transform property.

I'm confused by your PR -- to accomplish what you want, wouldn't the thing to do be `delete rec.transform`?

@zertosh and I are discussing whether a property for file rows (called `transform` or whatever) makes sense to allow opting a file out of transforms. See #1057.

zoubin:
@jmm Delete `rec.transform` would be just all right. I am closing this pr.

PR #1195, opened 2015-04-03
Fix #1194

### LINKED ISSUE #1194: 9.0.6 : Cannot catch errors + have to call bundle() twice
Hey, 

I have 2 issues in my gulp + watchify + browserify setup with the latest browserify build. Reverting back to browserify 9.0.4 fixed the issues:

- after any watchify update, I have to call `bundle()` twice to get my changes in the generated bundle
- I can no longer catch bundle() errors (with `on('error')`)

Here is the relevant piece of my Gulpfile: https://gist.github.com/cblavier/887346babc2ff173dfaf

Thanks,
Christian

jcppman:
since 9.0.6, `bundle()` returns a read-only stream by `var output = readonly(this.pipeline)`

but if `this._bundled` is true, a 'reset' will be triggered, which cause the
problem mentioned in #1194, because `this.pipeline !== output` after `this.reset()`, they are not the same thing anymore.

### DIFF
--- a/index.js
+++ b/index.js
@@ -732,7 +732,6 @@ Browserify.prototype.reset = function (opts) {
 
 Browserify.prototype.bundle = function (cb) {
     var self = this;
-    var output = readonly(this.pipeline);
     if (cb && typeof cb === 'object') {
         throw new Error(
             'bundle() no longer accepts option arguments.\n'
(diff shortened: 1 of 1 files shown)


terinjokes:
LGTM. Possible to get a test in for this?

substack:
Published in 9.0.7 with a test case. Thanks for the patch!

Issue #850, opened 2014-08-04
Expose property doesn't work when requiring a module name


tleunen:
With the latest version (5.9.1), aliasing a module name doesn't work anymore. `browserify -r fs:browserify-fs` doesn't create the alias.
With an older version (4.2.3), an alias was created with the same command.

No issue with specifying a file instead of a module name.

tleunen:
With the example, I still get the same error as before, with latest version. Is it something browserify won't support anymore? @substack 
```
browserify -r fs:browserify-fs
Error: Cannot find module 'browserify-fs' from ..
```

slorber:
@tleunen I still have the problem too. The build fail only when using an alias/expose name

code-guru:
I also have this problem both in cli and in code using `b.require('backbone', {expose: 'bb'})` . I am using Browsify `8.1.1`
``` bash
> browserify -r backbone:bb -o test.js
```

``` bash
Error: Cannot find module 'bb' from 'd:\project'
```

badem4o:
+1

Using module name instead of module path is still broken.

syadykin:
+1

Same here, import by name doesn't work with expose feature.

substack:
Published @ElNouch's fix in 9.0.8.

PR #1202, opened 2015-04-09
Fix expose field for external modules and builtins

ElNounch:
Added the 'file' field at row creation in require() for external modules and builtins.
Added two tests cases.
Unnested ternary operators (style only).
Fix current https://github.com/substack/node-browserify/issues/850
Feedback very welcome !


substack:
Looks great, thanks for the patch! Published in 9.0.8.

Issue #1170, opened 2015-03-17
When running browserify command, "Cannot find module" if using wildcards in Windows


awvalenti:
When I run on Windows:
* browserify *.js -o bundle.js

It says:
* Cannot find module 'C:\myproject\\*.js' from 'C:\myproject'

It also happens using \*\*/\*.js. It only happens on normal cmd, not on Git Bash, which simulates UNIX environment.

Seems to be the same problem as https://github.com/substack/node-browserify/issues/865. But I'm not using Azure environment.

nikku:
Have never seen wildcards working using the browserify commandline. List all files explicitly instead.

chad-autry:
I don't think listing files explicitly is an appropriate solution. In particular when building test bundles as mentioned in the hand book it is extremely handy.

Luckily, I think this is a fairly easy to solve problem. See my fork https://github.com/chad-autry/node-browserify/commit/9ff7c55cc67a7ddbc64f8e7270bcd75fcc72ce2f

That commit fixes things to me, though I am a bit lost on how to create a test case for it.

traviscooper:
+1

franleplant:
+1 for this! What is the state of the PR?

substack:
It doesn't sound like it should be browserify's job to solve missing features in particular shells. There are thousands of shells on hundreds of platforms.

franleplant:
@substack perhaps it could be nice to have a glob-cli that homogenizes those difference and improves portability and composition?



```shell
glob *.js | browserify -o output.js

#or perhaps

browserify (glob '*.js') -o output.js
```

I dont know Im just throwing an idea :)

nmn:
I don't think browserify needs to support this feature.

chad-autry:
So making a bit more noise,

I believe the changes I made to the pull request address the concerns about the code itself originally not being a good fit for browserify. There were no more follow up comments after the initial issues, I'm guessing it hasn't been re-reviewed.

So far as whether this change is a good fit for Browserify or not, the Browserify Handbook is currently incorrect. It states under the testing section

> Putting together all these steps, we can configure package.json with a test script:
>
>{
>  "name": "hundreder",
>  "version": "1.0.0",
>  "main": "index.js",
>  "devDependencies": {
>    "tape": "^2.13.1",
 >   "testling": "^1.6.1"
 > },
 > "scripts": {
 >   "test": "tape test/*.js",
 >   "test-browser": "browserify test/*.js | testlingify"
 > }
[...]

PR #1205, opened 2015-04-10
Accept glob string for [entry files] ...

chad-autry:
...in command "browserify [entry files] {OPTIONS}"
Windows does not auto expand globs on the command line

Creating the pull request to try and progress. It works locally for me, but an automated test case would be awesome. I can't tell how to make such a test case examining the existing suite.

This pull request is for https://github.com/substack/node-browserify/issues/1170

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,6 +15,7 @@ var inherits = require('inherits');
 var EventEmitter = require('events').EventEmitter;
 var xtend = require('xtend');
 var copy = require('shallow-copy');
+var glob = require('glob');
 var isarray = require('isarray');
 var defined = require('defined');
 var has = require('has');
@@ -39,6 +40,11 @@ function Browserify (files, opts) {
         opts = xtend(opts, { entries: [].concat(opts.entries || [], files) });
     }
     else opts = xtend(files, opts);
+    
+    if (isarray(opts.entries) && opts.entries.length === 1 
+      && typeof opts.entries[0] === 'string' && glob.hasMagic(opts.entries[0])) {
+       opts.entries = glob.sync(opts.entries[0]);
+    }
     
     self._options = opts;
     if (opts.noparse) opts.noParse = opts.noparse;


terinjokes:
I'm reviewing

terinjokes:
I think something similar the following would be prefered:

```javascript
[].concat(opts.entries).filter(Boolean).reduce(function (globbed, input) {
  return globbed.concat(glob.sync(input));
}, []).forEach(function (file) {
  self.add(file, { basedir: opts.basedir });
});
```

zertosh:
Ah this would work great in watchify (see https://github.com/substack/watchify/issues/56). Don't forget to test it with the CLI not just the API :stuck_out_tongue:

chad-autry:
@terinjokes Yeah, I see how that fits more in with the current pattern. I just need to spend some more time to understand the current pattern a bit better. I'll try to at least get around copy pasting your suggested change and testing it.

@zertosh That issue does not appear to be the same thing. That issue is about NEW files. This issue is about globs simply not working at all (on Windows). Solving this problem in browserify did seem to solve the duplicate problem of globs not working with watchify when I pointed it at my edited repo.

zertosh:
@chad-autry My is to use the fact that `entries` maybe be globs to monitor for new files when implementing it in watchify - that's all.

substack:
I don't know if this should be browserify's responsibility in the API. People can always call `glob()` themselves here.

chad-autry:
@zertosh You wouldn't be able to since Linux systems would still expand the glob before it reached watchify

@substack I'll try to re-write it like the other other uses of glob in bin/args.js to not do glob.sync()  For my use case I'm calling browserify from an nm script, was unable to call another program to substitute command parameters.

zertosh:
@chad-autry Not if it's quoted.

The use case I see in watchify is to monitor a directory of tests - where each test file needs to be an entry.

Watchify would look at `b._options.entries` to see if anything passes `glob.hasMagic`. If that's true, then watchify would setup another watcher with that glob, and on add/remove, it'll simply teardown the existing browserify instance and create a new one. This would only work with the CLI, since you have control there to create / destroy browserify instances.

Though, now that I think about it some more, this seems like a bad idea since it'll leave API users out in the cold.

chad-autry:
I've made the logic more consistent with how browserify does things, and moved it over into "bin/args.js" as @terinjokes suggested.

Lucas-C:
What is the status of this PR ? Can it be merged ?

chad-autry:
Requires re-review and hopefully acceptance by someone with permission to merge.

nemoxps:
@chad-autry Thank you so much for this patch! You saved my day :)
I can't believe this isn't merged yet. That is just poor. Probably going to use Webpack for further projects...

mukururo:
Let's have the best globally moving projects.

3cp:
Please consider this improvement. It's very useful to run tests.

```bash
browserify 'test/**/*.spec.js' | tape-run
```

PR #1210, opened 2015-04-12
Note that opts.standalone will be camelized on window if it's hyphenated

bendrucker:


jmm:
@bendrucker Thanks for the PR. I agree those name coercions should be documented, but they happen in https://github.com/ForbesLindesay/umd. It may be better to just link to documentation on it there. In any case, I don't think there *is* documentation of it there, so even if it makes sense to spell out the rules in the browserify documentation, I think it should *at least* be documented in the home project first. I'm actually not sure why it does so much coercion. I think until pretty recently it stripped `$` out of module names too.

How do you feel about submitting a PR there to document this stuff so we could link to it from here?

bendrucker:
Sure, happy to do PRs for docs and tests on umd for the name stuff. I'm aware of the umd dep and actually only discovered the camel casing behavior when I thought I'd need to do it myself but went to double check the `-s` output. 

I see the potential for frustration if you actually wanted hyphens raised by @Swaagie in #1212. I'm publishing modules that are often hyphenated as UMD builds on tags (https://github.com/bendrucker/publicist) to (begrudgingly) support Bower/copy-paste users. Not sure when anyone would ever want to do `window['my-package']`. I think it's reasonable to try to only output valid JS identifiers.

jmm:
> Sure, happy to do PRs for docs and tests on umd for the name stuff.

Awesome, thanks!

> I'm aware of the umd dep and actually only discovered the camel casing behavior when I thought I'd need to do it myself but went to double check the -s output. 

Yeah, I don't really get why it coerces the names and doesn't document it.

> Not sure when anyone would ever want to do `window['my-package']`.

I'm not sure either, I just know that if I don't meddle with the ability to do it I don't need to worry about it, and that if and when I want to do it for whatever reason I'll be able to. And case in point, we know @Swaagie wants it.

>  I think it's reasonable to try to only output valid JS identifiers. 

Not sure I agree with that, unless maybe it's a constraint of one of the module systems that makes sense to apply across the board for consistency. Otherwise I'm not sure what compelling reason there'd be to deviate from normal JS rules: if you want a `-` or other non-identifier character in a property name you can quote it and use brackets. So if you pass such a name as `standalone`, access it accordingly.

Swaagie:
It's more that I stumbled on it by accident. But suppose package is part of an automated system, the system would then loose reference to the build/exposed object as the name magically changed. Valid JS identifiers still applies if it is part of an object, it does not really matter if you access `window['nameWithSpecialChars']` or `nameWithSpecialChars` cause it is in scope.

zertosh:
@bendrucker If you add a section to the readme in [ForbesLindesay/umd](https://github.com/ForbesLindesay/umd) that describes this behavior, so it can referenced from here, I'd be more than happy to merge it. Just mention me.

bendrucker:
Updated to link to umd docs added in https://github.com/ForbesLindesay/umd/pull/23

zertosh:
:+1:

zertosh:
@terinjokes One more :smile:

PR #1211, opened 2015-04-12
Sanitize json modules

zertosh:
Not all valid JSON is valid JS. http://timelessrepo.com/json-isnt-a-javascript-subset


terinjokes:
Is there any reason why we shouldn't just replace these four characters?

zertosh:
@terinjokes Modularize everything! But sure. I could just inline the regexp if you want.

terinjokes:
Sorry, the module name was throwing me off. :~)

zertosh:
@terinjokes Get this one out too?

PR #1218, opened 2015-04-17
Array support for b.exclude() and b.ignore()

bakape:
Simple usability enhancement to prevent long exclude/ignore chains.

### DIFF
--- a/test/exclude.js
+++ b/test/exclude.js
@@ -1,0 +1,21 @@
+var browserify = require('../');
+var test = require('tap').test;
+var vm = require('vm');
+
+test('exclude array', function(t) {
+	t.plan(2);
+
+	var b = browserify();
+	b.add(__dirname + '/exclude/array.js');
+	b.exclude([
+		__dirname + '/exclude/skip.js',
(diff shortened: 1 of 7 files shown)


terinjokes:
Can you add tests for this behavior?

bakape:
Correct me, if I'm missing something, but is a test necessary here? The two functions simply recurse with arrays, so a test would involve mimicking the existing tests in a chain. A bit redundant. Of course, if you meant exactly that, no reason it can't be done.

christophercrouzet:
Any news on this one? I wouldn't mind having this merged as it currently is a bit inconsistent with other methods such as `external`, on top of having no idea how to programmatically define a variable number of `exclude`.

bakape:
@christophercrouzet 
Tests added.

christophercrouzet:
Thanks!

terinjokes:
@bakape for the record, I asked for tests not because it was difficult or tricky, but because we don't want to accidentally break it in the future.

This PR looks reasonable. Let's see what another contributor thinks.

PR #1222, opened 2015-04-19
Remove 'v' shortcut for version

zertosh:
This is part of an effort to bring consistency to the browserify and watchify flags (as somewhat discussed in https://github.com/substack/watchify/pull/204#issuecomment-93759292).

watchify already uses "v" for "verbose", and has no "version" flag. By removing the (arguably) seldom used "v" from browserify, (1) watchify can fully "own" it, and (2) can add a "version" flag without clashing with browserify's short form of it. Also, a lot of unix tools don't have a "v", just "version".

Alternatively, "v" could be replaced with "V" - [ssh](http://linux.die.net/man/1/ssh), [less](http://linux.die.net/man/1/less), among other do this.

cc: @jmm

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -20,7 +20,7 @@ if (b.argv._[0] === 'help' || b.argv.h || b.argv.help
         .on('close', function () { process.exit(1) })
     ;
 }
-if (b.argv.v || b.argv.version) {
+if (b.argv.version) {
     return console.log(require('../package.json').version);
 }
 


terinjokes:
What's the use case of having watchify be verbose without browserify also being verbose? From my perspective, I'd want my entire toolchain to be verbose while debugging, not just a portion of it.

zertosh:
@terinjokes Seems weird to have browserify be verbose when its output goes to stdout - but sure, maybe verbose logging should go to stderr?

FYI: watchify's verbose is simply "ZZZ bytes written to YYY.js (X seconds)" every time you rebuild. It's not really debugging. It's more like informational since it's a long running process.

zertosh:
@terinjokes @jmm Can we merge this in?

jmm:
@zertosh Sorry for delayed reply. I have no problem with it personally, but as I mentioned on another issue, I'm no authority on the CLI. So I think someone who has more interest / expertise with it should make the call.

terinjokes:
What do you mean by " watchify can fully "own" it"?

zertosh:
@terinjokes I mean that `v` only does what watchify already does with it.

Right now, the flag matrix looks like this:

| flag  | browserify | watchify |
| ----- | ---------- | -------- |
| `version` | version | noop |
| `v` | version | verbose |
| `verbose` | noop | verbose |

I think it makes more sense like this:

| flag  | browserify | watchify |
| ----- | ---------- | -------- |
| `version` | version | version
| `v` | noop | verbose |
| `verbose` | noop | verbose |

If browserify were to implement some kind of "verbose", it would then look like this:

| flag  | browserify | watchify |
| ----- | ---------- | -------- |
| `version` | version | version
| `v` | verbose | verbose |
| `verbose` | verbose | verbose |

zertosh:
(wow charts look pretty cool)

zertosh:
@terinjokes I'd really appreciate an answer on this. A version flag might seem trivial for watchify, but I get a lot of issues like https://github.com/substack/watchify/issues/190#issuecomment-94943128 - where users don't know whether they're using a local/global version and what version of browserify their watchify is using. The version flag I want to implement in watchify would answer those two questions (see https://github.com/substack/watchify/pull/197#issuecomment-94240851).

Without the change, if I add the version flag to watchify, we'd be looking at:

| flag  | browserify | watchify |
| ----- | ---------- | -------- |
| `version` | version | version |
| `v` | **version** | **verbose** |
| `verbose` | noop | verbose |

Maybe I'm over thinking this. If you agree, then let me know so I can proceed.

terinjokes:
Merged, sorry for the delay.

zertosh:
@terinjokes No worries! It's cool. Thanks!

Issue #1220, opened 2015-04-17
Constructor `opts.require` (keep, document, fix?; kill?)


jmm:
The constructor currently accepts an undocumented `opts.require` prop. I assume it's a holdover from an earlier version? In any case, can it be either documented and fixed or dropped? Personally I like the idea of having it, if there's not some good reason not to.

It's somewhat broken currently:

```js
var
  browserify = require('browserify'),
  fs = require('fs'),
  files = [{file: './a'}];

browserify()
  .require(files)
  .bundle()
  .pipe(fs.createWriteStream('./bundle-a1.js').on('finish', function () {
    browserify({require: files})
      .bundle()
      .pipe(fs.createWriteStream('./bundle-a2.js'));
  }));
```

The first produces a good `bundle-a1.js`, the second produces `TypeError: Arguments to path.resolve must be strings` and empty `bundle-a2.js`.

terinjokes:
Does this work if you clone files before being passed into each `.require`. I don't immediately see a difference between in the code paths.

jmm:
The difference is that in the first case it loops `files` and passes and calls `.require()` again with `files[0].file` as the first arg, while in the second case it loops `files` and calls `.require()` with `files[0]` as the first arg.

terinjokes:
It's concerning to me that `b.require({file: './a'});` has different behavior than `b.require([{file: './a'}]);`.

Resolving this difference seems like it would make the `browserify({require: [{file: './a'}]})` behavior work properly.

jmm:
@terinjokes

> It's concerning to me that `b.require({file: './a'});` has different behavior than `b.require([{file: './a'}]);`.

I ran into that between the time I posted and read your reply, and I agree. It surprised me.

> Resolving this difference seems like it would make the `browserify({require: [{file: './a'}]}) `behavior work properly.

That wouldn't totally resolve it on its own, but you could then resolve it by just calling `self.require(opts.require || [])` in the constructor. Actually currently the constructor does `(opts.require).filter(Boolean)`, but that's probably better handled in `b.require()` anyway.

goodbomb:
Hi, just to chime in quickly, I'm running into this problem right now and it's a real issue for me. I get this error no matter what I try:

```Error: path must be a string```

jmm:
[...]

PR #1224, opened 2015-04-21
readme: document opts.transform

LinusU:


zertosh:
Thanks @LinusU, but I think they're a little more than "transform functions", and "modules" should be singular. How about borrowing some the language from https://github.com/substack/node-browserify#btransformtr-opts?

LinusU:
Yes I was struggling a bit with what to write. It seems redundant to document how they works in two places in the documentation...

Do you think that we could link to `b.transform(tr, opts)` or maybe create a new section documenting how the source transforms works, and then linking to that?

zertosh:
Yeah something short that links to https://github.com/substack/node-browserify#btransformtr-opts would be cool. The "Usage" section has a quick one-liner for "transform".

jmm:
I've been thinking that eventually it might be a good idea to have a glossary kind of section for defining things like logical ideas and data structures that are referenced in multiple places. So you could have an entry for `transform` that's linked to from `opts.transform` (if that indeed gets documented) and `b.transform()`. That's what I was alluding to in #1226 when I talked about documenting `file object` and `entry file` at some point. And it'll be the same deal as here if `opts.require` gets documented as discussed in #1220.

mattfysh:
Is there a reason for this PR not being merged? we're hoping to use this API but not too comfortable with its undocumented state

LinusU:
Thanks!

Issue #707, opened 2014-03-27
Reducing the size of browserify when installed


hughsk:
Right now, browserify's sitting at over 50MB to install through npm. Though when running it through disc, it appears under 2MB of that is actually getting required:

![screen shot 2014-03-27 at 2 33 20 pm](https://cloud.githubusercontent.com/assets/569817/2533561/b10e063a-b560-11e3-85cf-b2dc5184f296.png)

I ran a fresh install through DaisyDisk:

![screen shot 2014-03-27 at 2 35 23 pm](https://cloud.githubusercontent.com/assets/569817/2533569/e0d75c72-b560-11e3-9e67-0909eb953a7d.png)

And there's a few modules including more test code than browserify itself needs to run, e.g. the url package alone being the worst offender at 20MB because punycode's got a copy of requirejs in there somewhere.

It might be worth looking into this for the sake of reducing `npm install` time and saving on disk space. I know I have tons of projects with local copies of browserify and it's starting to take its toll :)

Thoughts?

terinjokes:
Browserify has been using the 0.10.x version of `url/` since 9cd951a1994dddc9acf52f5124ae0e7f5466a1ec

aearly:
A lot of the filesize comes from test resources:

```
$ find node_modules/browserify/ -type d -name test | xargs du -s | sort -n
...
336	node_modules/browserify/node_modules/browser-resolve/test
388	node_modules/browserify/node_modules/crypto-browserify/node_modules/browserify-aes/test
400	node_modules/browserify/node_modules/resolve/test
448	node_modules/browserify/node_modules/insert-module-globals/node_modules/JSONStream/test
448	node_modules/browserify/node_modules/module-deps/node_modules/JSONStream/test
452	node_modules/browserify/node_modules/browser-pack/node_modules/JSONStream/test
452	node_modules/browserify/node_modules/deps-sort/node_modules/JSONStream/test
456	node_modules/browserify/node_modules/JSONStream/test
528	node_modules/browserify/node_modules/module-deps/test
900	node_modules/browserify/node_modules/module-deps/node_modules/detective/node_modules/escodegen/node_modules/esprima/test
1572	node_modules/browserify/test
1756	node_modules/browserify/node_modules/insert-module-globals/node_modules/lexical-scope/node_modules/astw/node_modules/esprima-fb/test
2408	node_modules/browserify/node_modules/module-deps/node_modules/detective/node_modules/acorn/test
[...]

PR #1138, opened 2015-02-24
Factor our repetitive transform setup code

jmm:
Next commit adjusts some formatting to be more consistent with the rest of the file (`function (` vs `function(`).

PR #1230, opened 2015-04-26
Update deps

zertosh:
* The update of defined and xtend are so we can get npm to dedupe them. It's already in https://github.com/substack/insert-module-globals/pull/42, https://github.com/substack/deps-sort/pull/4, https://github.com/substack/module-deps/pull/82 and https://github.com/substack/browser-pack/pull/56.

* I'm also updating and pinning punycode to 1.3.2 for deduping purposes. But I chose to not make it floating since [url](https://github.com/defunctzombie/node-url/blob/v0.10.3/package.json#L6) also has it pinned there.

* And I'm updating module-deps, despite that the `^` already covers the version, so that over in watchify I can count on https://github.com/substack/module-deps/pull/80 always being in whatever is the next browserify version we publish.

The motivation for the deduping is https://github.com/substack/node-browserify/issues/707#issuecomment-96273418.

---

@jmm and @terinjokes: This should be published as 9.0.9. But since https://github.com/substack/node-browserify/pull/1222 hasn't been published, strict semver dictates that we move to 10.0.0. I think that's over kill for removing the "v" flag. I'll hold off publishing until you chime in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "insert-module-globals": "^6.2.0",
     "isarray": "0.0.1",
     "labeled-stream-splicer": "^1.0.0",
-    "module-deps": "^3.7.0",
+    "module-deps": "^3.7.7",
     "os-browserify": "~0.1.1",
     "parents": "^1.0.1",
     "path-browserify": "~0.0.0",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "constants-browserify": "~0.0.1",
     "crypto-browserify": "^3.0.0",
     "deep-equal": "^1.0.0",
-    "defined": "~0.0.0",
+    "defined": "^1.0.0",
     "deps-sort": "^1.3.5",
     "domain-browser": "~1.1.0",
     "duplexer2": "~0.0.2",
@@ -51,7 +51,7 @@
     "parents": "^1.0.1",
     "path-browserify": "~0.0.0",
     "process": "^0.10.0",
-    "punycode": "~1.2.3",
+    "punycode": "^1.3.2",
     "querystring-es3": "~0.2.0",
     "read-only-stream": "^1.1.1",
     "readable-stream": "^1.1.13",
@@ -69,7 +69,7 @@
     "url": "~0.10.1",
     "util": "~0.10.1",
     "vm-browserify": "~0.0.1",
-    "xtend": "^3.0.0"
+    "xtend": "^4.0.0"
   },
   "devDependencies": {
     "backbone": "~0.9.2",


terinjokes:
LGTM :+1: @jmm?

jmm:
This all sounds good. Having to pin a dep like punycode because it's pinned in another dep seems a little wonky though. Then it'll get out of sync if url updates it and no one thinks to update it here. npm doesn't provide a more elegant way to accomodate that situation, eh?

Re: the new version number, I have a bit of mixed feelings about that, since it probably wouldn't break 99.9% of things, but ultimately I guess I'd err on the side of caution / predictability and bump the major unless this stuff can wait until some other things that would necessitate that anyway get rolled in. I think there's some urgency for you to get this to fix stuff in watchify though, right @zertosh? At least it would be a super easy changelog to digest.

BTW, FWIW I'm still operating on the guidelines I was pointed to [here](https://github.com/substack/node-browserify/pull/1138) that "Declaring formal releases remains the prerogative of the project maintainer." -- therefore not for me to say :)

@terinjokes were you in favor of just staying at v9?

terinjokes:
@jmm With the process changes, I'll be in favor of a v10 (since it seems like it could break things subtly.

And true, I believe I've only published patch releases before.

jmm:
@terinjokes Yeah, sorry, I just got to that one (reading the whole thing in a bit) and saw that it might be a moot point.

> And true, I believe I've only published patch releases before.

I figured that, at least as far as it pertains to me at this point in time, that meant any release, even patch level. What would you say that means by "formal release" -- major only? major.minor? everything?

zertosh:
> . I think there's some urgency for you to get this to fix stuff in watchify though

@jmm Not urgent, just a convenience. But since the `process` update is kind of important, I figured this should ride along.

> "Declaring formal releases remains the prerogative of the project maintainer." 

O_O eep. I don't know. Is formal = major? Or does this mean that publishing rights to npm may or may not be given a contributor?

> The multi-stage install that npm@3 brings should dedupe a bit more than the current install. I'll have to do some tests to see if it would figure this one out.

@terinjokes When that time comes we should totally revisit this. We should be mindful of changes to `url` nonetheless.

terinjokes:
Unless we have clarification, probably best to allow substack to tag the releases then.

zertosh:
Makes sense to wait then - don't want to overstep bounds w/o clarification.

jmm:
I don't mean to screw up anyone else's workflow -- I assumed that people who've been collaborators here longer than me might have different parameters for how to contribute. (I think I was the most recent addition until you, @zertosh.)

Or maybe you're on to something here:

> Or does this mean that publishing rights to npm may or may not be given a contributor?

Maybe being designated an npm maintainer is implied thumbs up for cutting releases?

@terinjokes
> The multi-stage install that npm@3 brings should dedupe a bit more than the current install. I'll have to do some tests to see if it would figure this one out.

@zertosh
> @terinjokes When that time comes we should totally revisit this. We should be mindful of changes to url nonetheless.

I was thinking about that too, but I was wondering if it'll also dedupe on `npm update`. I just asked over here (which only mentions `npm install`): https://github.com/npm/npm/issues/6912#issuecomment-96818083. If that's the case, and once it arrives, then I think it'd make sense to depend on `~1.3.2` here (or whatever is the relevant version at the time) and let url dictate the exact version via npm. Or maybe even now, unless someone is really going to be keeping an eye on url. Is it even likely to cause a problem if they depend on different minor / patch versions?

terinjokes:
Having thought about this over the day, I'm not sure I like pinning to punycode here, since it creates a maintainability issue for us to remember to upgrade it when we upgrade url. @zertosh this is to save a few megabytes on install, correct?

zertosh:
> this is to save a few megabytes on install, correct?

@terinjokes correct. Also, it's not like `url` is getting it's `punycode` - when you `require('url');` it actually gets the `punycode` that's specified in builtins. So, the version specified by `url` itself doesn't matter for browserifying purposes.

zertosh:
> Maybe being designated an npm maintainer is implied thumbs up for cutting releases?

@terinjokes I'm starting to think that. hughsk's policy over at envify is "In the future I'd say submit a PR, and if you don't hear from me after a day then merge away :)" https://github.com/hughsk/envify/pull/23#issuecomment-78418310. And he gave me the same collaborator agreement.

terinjokes:
@zertosh I've felt comfortable merging in about the same time period. It's cutting releases that I'm not sure of.

substack:
All of this looks great, you don't need my permission to cut a release!

terinjokes:
With that out of the way, to the issue at hand.

I continue to prefer not pinning punycode, even if it does mean two get installed until user switches over to npm@3. As an aside, I'm not even sure why punycode is listed as a dependency since recent node versions will never resolve to it (but that's an issue for that repo).

Ignoring the questionable change to process.nextTick, do we have any other PRs pending (or about to be pending) that would require the bump?

zertosh:
@terinjokes I think that's it for now. I went ahead and made punycode ^1.3.2. I'll merge this in and https://github.com/substack/node-browserify/pull/1231. You can do the honors of cutting 10.0 :smile:

terinjokes:
@zertosh ok, happening now.

Issue #1179, opened 2015-03-24
Q stops working with Browserify >= 8.1.0 due to nextTick changes


slorber:
Q works fine with Browserify 8.0.3 but has serious problems since 8.1.0.
This is related to changes done in nextTick implementation of node-process: https://github.com/substack/node-browserify/pull/1018

Related Q issue: https://github.com/kriskowal/q/issues/669

I don't know who of Q or Browserify should fix this but this makes Q unreliable in production.

slorber:
A PR is available here https://github.com/defunctzombie/node-process/pull/37

slorber:
@substack @defunctzombie this can now be solved by uprading "process" dependency to 0.11 right?

zertosh:
@slorber It's not just flipping the version. A couple of tests fail for the reasons in https://github.com/substack/insert-module-globals/pull/42 (under process to ^0.10.0). I'll try to get to it tmw.

PR #1231, opened 2015-04-27
Update process to ~0.11.0

zertosh:
Fixes https://github.com/substack/node-browserify/issues/1179. This also solves my major-version-bump dilemma from https://github.com/substack/node-browserify/pull/1230 because this merits it for sure.

The reason for having to pass in `clearTimeout` into the context in tests is because there is now a queue in `process.nextTick` that depends on it (see https://github.com/defunctzombie/node-process/pull/38 and https://github.com/defunctzombie/node-process/compare/v0.10.1...v0.11.0.).

cc: @jmm @terinjokes

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "os-browserify": "~0.1.1",
     "parents": "^1.0.1",
     "path-browserify": "~0.0.0",
-    "process": "^0.10.0",
+    "process": "~0.11.0",
     "punycode": "^1.3.2",
     "querystring-es3": "~0.2.0",
     "read-only-stream": "^1.1.1",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "http-browserify": "^1.4.0",
     "https-browserify": "~0.0.0",
     "inherits": "~2.0.1",
-    "insert-module-globals": "^6.2.0",
+    "insert-module-globals": "^6.4.0",
     "isarray": "0.0.1",
     "labeled-stream-splicer": "^1.0.0",
     "module-deps": "^3.7.7",
--- a/test/bundle.js
+++ b/test/bundle.js
@@ -13,6 +13,7 @@ test('bundle', function (t) {
         
         var c = {
             setTimeout : setTimeout,
+            clearTimeout : clearTimeout,
             console : console
         };
         vm.runInNewContext(src, c);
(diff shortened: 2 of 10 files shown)


zertosh:
cc: @slorber

terinjokes:
Glad to get the Q issue resolved. I'm happy to merge both these in and do a major release. @jmm?

zertosh:
@terinjokes Cool! I merged https://github.com/substack/insert-module-globals/issues/43 and bumped it here in https://github.com/substack/node-browserify/commit/064f5a157995a3dbda14f04e1a8cf7120775afbe - just for consistency, doesn't really affect anything in browserify.

slorber:
nice is there a browserify release yet?

terinjokes:
@slorber not yet :sparkles:

jmm:
@zertosh @terinjokes You guys probably understand this issue better than I do, but I looked through a bunch of the related issues / PRs and it LGTM.

> This also solves my major-version-bump dilemma from #1230 because this merits it for sure.

Does this break BC somehow? I got the impression that stuff that's already working would continue working with this change. Or would the changes in defunctzombie/node-process/pull/38 actually break something people were relying on before? The other stuff I saw in recent commits there didn't seem like it would be breaking. Sorry if I'm being dense here.

Aside: it sounds like maybe the issue with Q in particular has been addressed independently [there](https://github.com/kriskowal/q/issues/645#issuecomment-96288816) -- not that it has any bearing on this PR.

zertosh:
@jmm I would think this is a BC because it fundamentally changes how `process.nextTick` works by enforcing a queue that doesn't die on errors (plus, now it depends on `clearTimeout`, which broke our tests). Granted, it is fixing a bug, but a bug that might've been relied on.  Since this is such an obscure feature - does that kinda fall into the semver gray area? I don't know :( I could really use some guidance right now.

jmm:
@zertosh I'm not an expert on semver, so it does seem like a gray area to me. If you think people may have relied on the previous behavior then I think it would make sense to bump the major version. I wonder what node-process would've done with the version if it had already been >= 1.

terinjokes:
Reading more into the issue for the changelog, I'll find it really strange if anyone *did* rely on the behavior, since it prevented any future process.nextTick tasks from running.

zertosh:
@terinjokes Maybe? I'm by no means a semver expert :stuck_out_tongue:

PR #1239, opened 2015-05-02
Fixed #899 - browserify removes output file if it didn't exist previously

### LINKED ISSUE #899: Browserify should not create empty output files on error
When an error occurs, Browserify 5.11.2 creates an empty output file.  This should not happen, as it messes up `make`-based workflows.

```
$ cat >foo.js <<EOF
var quux = require('quux');
EOF

$ browserify foo.js -o bar.js
Error: Cannot find module 'quux' from '/private/tmp/foo'
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:50:17
    at process (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:119:43)
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:128:21
    at load (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:60:43)
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:66:22
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:21:47
    at Object.oncomplete (fs.js:107:15)

$ ls -l | grep bar.js
-rw-r--r--  1 mietek  wheel   0 13 Sep 14:53 bar.js
```

kika:
The change is trivial. I didn't use fs.exists() because it's being deprecated soon according to the recent Node documentation.

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -56,7 +56,12 @@ var bundle = b.bundle();
 bundle.on('error', errorExit);
 
 var outfile = b.argv.o || b.argv.outfile;
+var outfileExists = true;
+
 if (outfile) {
+    var tmp = fs.createReadStream(outfile);
+    tmp.on('error', function() { outfileExists = false; });
+    tmp.close();
     bundle.pipe(fs.createWriteStream(outfile));
 }
 else {
@@ -77,6 +82,9 @@ function errorExit(err) {
     }
     else {
         console.error(String(err));
+    }
+    if (!outfileExists) {
+        fs.unlinkSync(outfile);
     }
     process.exit(1);
 }
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -56,12 +56,11 @@ var bundle = b.bundle();
 bundle.on('error', errorExit);
 
 var outfile = b.argv.o || b.argv.outfile;
-var outfileExists = true;
+var outfileExists = false;
 
 if (outfile) {
-    var tmp = fs.createReadStream(outfile);
-    tmp.on('error', function() { outfileExists = false; });
-    tmp.destroy();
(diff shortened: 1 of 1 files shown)


zertosh:
Why the tmp stream? Instead of:

```js
try {  fs.unlinkSync(outfile); } catch(err) { }
```

kika:
Generic answer : I personally hate software which does what it wasn't asked to do :-)
Specific answer: if, for whatever reason, the output file did exist before the browserify run it will get deleted without advance warning. So I check if the file exists before attempt to run and clean up after the failed run only if there was no output file before.

zertosh:
Ah makes sense. The `fs.createWriteStream(outfile)` will write to the file (thus emptying it if it existed before) pretty much always before any error ever gets emitted by browserify. But at least you're leaving the file behind. Ok cool, but that tmp stream :scream_cat: 

```js
var outfileExists;

if (outfile) {
    try { outfileExists = !!fs.lstatSync(outfile); } catch (err) {}
    bundle.pipe(fs.createWriteStream(outfile));
}
```

kika:
makes sense, thanks. Updated.

substack:
I'm against adding `.sync` calls. This patch also seems to add a lot more complexity that will probably break in strange ways. Writing to a tmp file was removed from watchify for this reason.

kika:
@substack, are you against this PR entirely, or just against replacing the test for existence by opening the stream with `lstatSync()`? Without this PR (or equivalent) browserify breaks `make`-based workflows and forces use of `make` kludges like `DELETE_ON_ERROR` which do not exist in all flavors of `make`

goto-bus-stop:
It looks like #899 was addressed by #1673. Thanks!

PR #1240, opened 2015-05-03
--no-browser-field and --node

substack:
This patch adds support for `--no-browser-field` to turn off the package.json browser field machinery. This is useful if you are targeting a node environment with the bundle, such as [code coverage](https://npmjs.com/package/coverify), to apply transforms to server code, or to target other platforms like chrome apps.

This feature is also available as `opts.browserField = false`. `--node` is added as an alias for `--bare` and `--no-browser-field`.


terinjokes:
:sparkles: LGTM

PR #1244, opened 2015-05-06
Remove unnecessary "isDedupe" json check

zertosh:
watchify doesn't cache when "deps" are emitted, it caches module-deps' output - which would be the preprocessed json.


zertosh:
@terinjokes @jmm: Any thoughts on this? Basically, this is a hack for users on watchify 2.4.0 and under.

PR #1245, opened 2015-05-06
Bump module-deps to ^3.7.10

zertosh:
```
# 10.1.1

* Updates module-deps to ^3.7.10, which fixes an issue where under certain conditions the directory for a `package.json` would be attributed to the wrong package. [substack/module-deps#81](https://github.com/substack/module-deps/pull/81) (thanks @pmowrer)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "insert-module-globals": "^6.4.0",
     "isarray": "0.0.1",
     "labeled-stream-splicer": "^1.0.0",
-    "module-deps": "^3.7.7",
+    "module-deps": "^3.7.10",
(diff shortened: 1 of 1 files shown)


substack:
The module-deps dependency is already set to `^3.7.7` so that seems unnecessary?

zertosh:
It's mostly for the changelog entry. What's a good way to let users know that there were relevant upstream fixes?

substack:
I think it's best if the relevant package (module-deps) handles that. Otherwise there's so much noise in the main project.

zertosh:
Ok cool. I'll start a changelog there.

substack:
Sounds good!

Issue #1219, opened 2015-04-17
CLI path resolution inconsistent for entry files and -e|--entry vs. -r|--require


jmm:
In browserify 9.0.8 CLI on linux path arguments are resolved inconsistently for entry files / `-e` / `--entry` vs `-r` / `--required`.  Seems counterintuitive. Is there a reason to resolve them differently? Note: triggering it doesn't require an `expose` argument for the `-r` files.

```
# Module files a.js and b.js

# Error: Cannot find module 'b'
browserify a -r b
browserify -e a -r b

# No error, both bundled
browserify a -r ./b
browserify -e a -r ./b
```

terinjokes:
I believe `entry` is really expected to be a file, whereas `require` is documented as being anything that can be resolved by the node require resolution. I'm not aware of any history for them doing different things.

That said, this is method is kinda gnarly, and I should really take some time to clean it up.

jmm:
@terinjokes Thanks for the reply. I'm not sure I see the distinction -- shouldn't they both just be resolved to a module file starting from some base dir (the same one in the case of these examples)? Unless there's a good reason to do otherwise it seems like it would be ideal for them to perform consistently in terms of the input they can take and how they resolve it to a file. No?

terinjokes:
@jmm I was hoping our mysterious leader would come by and comment about why they do different things. :wink:

jmm:
@terinjokes Ah, got it, sounds good.

substack:
`--require` looks up files the `require.resolve()` way because that's what `require()` does.

`--entry` expects file paths like `fs.createReadStream()` would

It's difficult to do both operations in a single step. Sometimes you can `require()` a path but other times you can only `fs.createReadStream()` a path. Sometimes you can do both and sometimes neither one can work. When you do both, it's possible to have 2 different outcomes.

If you have these files:

* whatever.js
* node_modules/whatever.js/index.js

Then the string `"whatever.js"` passed to either option will give back different results.

jmm:
Thanks for explaining that @substack. I see what you guys mean, and that totally makes sense if the 80% case for `-e|--entry` is files outside of `node_modules`, and the opposite is the case for `-r|--require`. I don't know why I was so confused on this, even through your explanation @terinjokes.

Issue #1203, opened 2015-04-09
9.0.8 breaks 'dep' event


jwalton:
Since 'dep' isn't documented in the browserify readme, maybe this was intentional, but it's causing grief for me, so I thought I'd raise it in case it wasn't.  :)

In browserify 9.0.7, if I did:

    b = browserify(opts);
    b.on('dep', function(dep) {
        console.log(dep.id + " -- " + dep.file);
    });

then I'd get output like:

    backbone.layoutmanager -- /Users/jwalton/project/node_modules/backbone.layoutmanager/backbone.layoutmanager.js

In 9.0.8, I get:

    backbone.layoutmanager -- backbone.layoutmanager

`dep.file` is not actually a file here.  I'm using [uber-watchify](https://github.com/benbria/uber-watchify), which tries to `fs.statSync(dep.file)`, which is obviously going to end in tears for all concerned.

jwalton:
Looks like the 'deps' pipeline has the same problem:

    b.pipeline.get('deps').push(through.obj(function(row, end, next) {
        console.log(row.id + ' -- ' + row.file);
        next();
    }));

gives the same output for 9.0.7 and 9.0.8.

jwalton:
This is going to break watchify, too, although less catastrophically.  Watchify will likely end up watching the non-existent `process.cwd() + '/backbone.layoutmanager'` instead of the file it's supposed to be watching.  If this file gets created it will cause a rebuild.

jmm:
@jwalton Thanks for the report.

`dep` is mentioned in the readme in the pipeline section:

> 'emit-deps' - emit 'dep' event

#1202 is what changed. This is related to the issue I raised in #1162 -- `row.file` (or `rec.file`) properties don't have consistent values. Sometimes they're absolute paths and sometimes not (like when an object with a `file` prop is passed to `b.add()` or `b.require()`).

I mentioned that with respect to watchify in https://github.com/substack/watchify/pull/160 and @zertosh said it wouldn't matter in that case, but not sure if that's exactly the same situation you're talking about.

jwalton:
[...]

Issue #1162, opened 2015-03-12
What roles are rec.id and rec.file supposed to play?


jmm:
In browserify's public methods (like `b.add()` and `b.require()`) and the pipeline, what are the respective roles of `rec.id` and `rec.file` (or `row.id` and `row.file`) supposed to be? I'd like the clarify this, document what's appropriate, and if necessary / possible perhaps even simplify the logic around these props.

Throughout browserify and the pipeline they do this strange dance where sometimes they're used sort of interchangeably and other times not and their meanings can vary in confusing ways depending on input provided to browserify and pipeline phase. One explanation that could make sense is if `rec.id` is equivalent to the `id` in `require('id')` and `rec.file` is the resolved pathname, but they're not consistently used that way. `rec.id` is also involved in the logic for linking and exposing modules. Examples:

Suppose I have these files:

`entry.js`
```js
console.log(require('./dep').msg);
```

`dep.json`
```json
{"msg": "This is dep.json"}
```

* The `json` pipeline phase tests only the `file` value for `/\.json$/`.

* If I do `browserify('./entry').bundle()` I get a bundle that logs "This is dep.json".

  In this case module-deps parses the `require('./dep')` and pushes a row like `{id: '/abs/path/to/dep.json', file: '/abs/path/to/dep.json'}`. `json` pipeline phase recognizes it as JSON. 

* If I do `browserify().require('./dep', {expose: 'DEP'})` I get `SyntaxError: Unexpected token`.

  In this case `b.require` pushes a row like `{id: 'DEP', file: './dep'}`. Even though it has resolved the pathname, module-deps pushes a row with the same props. `json` pipeline phase is unable to recognize it as JSON.

* If I do `browserify().require({file: './dep', expose: 'DEP'})` I get `SyntaxError: Unexpected token`.

  `b.require()` pushes a row like `{file: './dep', id: {file: './dep'}}`. Even though it has resolved the pathname, module-deps pushes a row with the same props. `json` pipeline phase is unable to recognize it as JSON.

* If I do `browserify().require({id: './dep', expose: 'DEP'})` I get no error (and the `json` pipeline transform has been applied). Passing an object with an `id` prop is not documented, so take this one for what it's worth. But plugins may have an interest in pushing / modifying rows, so understanding what these props are for and how to use them throughout the pipeline is important. The browserify handbook includes an example of reading / writing `rec.id` during the `deps` / `label` phases.

[...]

PR #1248, opened 2015-05-06
Make sure entry paths are always full paths

zertosh:
When using the CLI, entry files were always resolved to absolute paths, but when using the API, they weren't. module-deps respects the `row.file` on its initial rows insofar that it resolved relative paths (based on `opts.basedir`) but it didn't rewrite it on the `row`. This led to a weird behavior with watchify where entry files were never being pulled from the cache - unless you used the CLI or passed in an absolute path via the API.

Using `basedir` to resolve the path is consistent with @substack's view on path resolution ("require vs entry" in https://github.com/substack/node-browserify/issues/1219#issuecomment-96126563).

This PR also adds a few tests that use the relative path behavior and extends others to verify that `row.file` is always the absolute path. 

cc: @jmm @terinjokes

### DIFF
--- a/bin/args.js
+++ b/bin/args.js
@@ -61,7 +61,7 @@ module.exports = function (args, opts) {
             s.resume();
             return rs;
         }
-        return path.resolve(process.cwd(), entry);
+        return entry;
     });
     
     if (argv.node) {
--- a/index.js
+++ b/index.js
@@ -180,7 +180,11 @@ Browserify.prototype.require = function (file, opts) {
         self._bpack.hasExports = true;
     }
     
-    if (row.entry) row.order = self._entryOrder ++;
+    if (row.entry) {
+        row.file = path.resolve(basedir, row.file);
+        row.order = self._entryOrder ++;
+    }
+    
     if (opts.transform === false) row.transform = false;
     self.pipeline.write(row);
     return self;
--- a/test/entry_relative.js
+++ b/test/entry_relative.js
@@ -2,10 +2,10 @@ var browserify = require('../');
 var vm = require('vm');
 var test = require('tap').test;
 
-test('entry', function (t) {
+test('entry - relative path', function (t) {
     t.plan(3);
-    
-    var b = browserify(__dirname + '/entry/main.js');
+
+    var b = browserify('entry/main.js');
     b.on('dep', function(row) {
         if (row.entry) t.equal(row.file, __dirname + '/entry/main.js');
     });
@@ -21,11 +21,11 @@ test('entry', function (t) {
     });
 });
 
-test('entry via add', function (t) {
+test('entry - relative path via add', function (t) {
     t.plan(3);
     
-    var b = browserify();
-    b.add(__dirname + '/entry/main.js');
+    var b = browserify({basedir: __dirname});
+    b.add('entry/main.js');
     b.on('dep', function(row) {
         if (row.entry) t.equal(row.file, __dirname + '/entry/main.js');
     });
--- a/test/multi_entry.js
+++ b/test/multi_entry.js
@@ -10,14 +10,20 @@ var testFiles = [
 ];
 
 test('multi entry', function (t) {
-    t.plan(3);
+    t.plan(6);
     
     var b = browserify([
         testFiles[0],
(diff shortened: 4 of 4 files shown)


substack:
Looks great, published in 10.1.1.

jmm:
@zertosh @substack I think we need to decide *exactly* how (and where) entry file paths will be resolved and document it more. This PR is a bit different, but reminiscent of the situation with #1033 / #1059 / #1077. This PR pseudo-resolves the paths and module-deps still actually resolves them, so unfortunately this does not ensure that `row.file` is an absolute pathname that actually exists or the same path that module-deps resolves.

Here's a very simple repro (given an `entry.js` in `cwd`):

```js
var
  browserify = require('browserify'),
  row_flow = require('browserify-row-flow');

browserify('./entry')
  .plugin(row_flow().plugin())
  .bundle()
;  
```

You can see that it exits the `record` phase (and `deps`) with `file: '/basedir/entry'` and is emitted by the `file` event with `file: '/basedir/entry.js'` (the value that module-deps actually resolved it to).

Re: https://github.com/substack/node-browserify/issues/1219#issuecomment-96126563, I almost brought this up over there that the CLI and API resolve entry file paths differently, but I figured it's a divergence that makes sense so I didn't bother. As of v10.1.0 entry files provided to the API *were* resolved using the module resolution algorithm. So the following were not equivalent:

```js
browserify('./entry');
browserify('entry');
```

I figure it makes sense for the CLI to resolve `entry` by appending it to the `basedir` because that's probably the 80% use case and character count is much more significant than for API calls. With the API, if you want entry paths appended to the `basedir` instead of put through the module resolution algorithm it's trivial to prepend `./`.

This PR changes the resolution behavior of the second case:

```js
browserify('entry');
```

Previously, if there were a `node_modules/entry/index.js`, this would've been resolved to it. Now it appends it to `basedir`.

zertosh:
Probably the worst behavior is having the API and CLI do different things. Sure, this PR doesn't ensure that the file exists, but it brings consistency. 

> I think we need to decide exactly how (and where) entry file paths will be resolved and document it more.

The best thing to do is to just delegate to mdeps. [This](https://github.com/substack/module-deps/blob/9be8dea5b21a905a1dd490124b167f42e5df95d0/index.js#L412) line is what preserves the input row's file. Just remove the `if` and all rows will have a `file` prop that is the mdeps resolved path. Want me to look into this?

> With the API, if you want entry paths appended to the basedir instead of put through the module resolution algorithm it's trivial to prepend ./

It's still trivial. Just pass in an absolute path :)

jmm:
> Probably the worst behavior is having the API and CLI do different things. Sure, this PR doesn't ensure that the file exists, but it brings consistency.

Maybe. They're pretty different animals though and I'd hate to see compromises made for the CLI pollute the API. The CLI and API would be more consistent if you could do `b.require('some_file.js:alias')`, but I doubt anyone thinks that's a good idea.

> The best thing to do is to just delegate to mdeps.

That's pretty much what I've been thinking and what I did for a similar case in #1077.

> This line is what preserves the input row's file. Just remove the `if` and all rows will have a `file` prop that is the mdeps resolved path. Want me to look into this?

For me it just circles back to my questions in #1162 -- what is `file` supposed to represent? Like I said in https://github.com/substack/node-browserify/issues/1203#issuecomment-91672659 my working theory is that it *should* "always be a string that represents an absolute pathname, perhaps a sham value for streams or other forms of in-memory rows" (or at least by the time module-deps is done with it it should). So if that really is what it should represent, then yes, making that assignment unconditional would be the thing to do.

> It's still trivial. Just pass in an absolute path :)

Well it's even more trivial now since it happens by default. What is now less trivial is using a module name to refer to an entry file. I guess now you'd have to manually invoke node-resolve on it, or if you're sure you want to resolve it from `cwd`, `require.resolve()`.

Gillingham:
It looks like this broke being able to name a module and file location using the browser field in package.json

jmm:
@Gillingham Can you please open an issue with a simple reproduction for the `browser` field part?

Gillingham:
@jmm opened #1250

PR #1253, opened 2015-05-08
Optimization: Add syntax check cache

zertosh:
This PR introduces a caching system for the syntax check. It's a big optimization for repeat builds (think watchify). I had proposed this first in https://github.com/substack/node-syntax-error/pull/9, but substack suggested that it was better implemented here.

I went with a per instance cache instead of a global cache (though it could very safely be one) because I don't want to set the precedent that it's ok for Browserify to be stateful.


zertosh:
@mattdesl Thanks for pointing this out!

terinjokes:
LGTM.

PR #1257, opened 2015-05-09
Fix globals transform noParse path matcher

zertosh:
* Fix globals transform noParse path matcher
    * Fixes passing in a function. The logic to call the function is there, but if you passed one in, `path.resolve` would throw because it wasn't a string.
    * Fixes resolving from the `basedir` - not assuming it's the cwd.
    * Optimizes a bit by moving `no` and `absno` into the closure so they're only built once per `b.bundle()`.
    * Adds some tests for this.

* Add test for "noParse"
    * Adds a test for the only way `noParse` for deps works - an array of full paths.

(related to https://github.com/substack/node-browserify/pull/1256)


zertosh:
@jmm @terinjokes Quick review?

Issue #1266, opened 2015-05-15
Dependency number is undefined for relative includes when specifying many files


ptarjan:
With this simple setup

```
$ cat a.js
var b = require('./b.js');
console.log(b.foo());
$ cat b.js
exports.foo = function() {
  return 'works';
};
```

Then with version 9.0.8 I happily get `"./b.js":2`

```
$ browserify --version
9.0.8
15:16:16 pt@Pauls-MBP:/tmp/browsesrify$ browserify a.js b.js
(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
var b = require('./b.js');
console.log(b.foo());

},{"./b.js":2}],2:[function(require,module,exports){
exports.foo = function() {
  return 'works';
};

},{}]},{},[1,2]);
```

But with 10.2.0

```
$ node_modules/browserify/bin/cmd.js --version
10.2.0
$ node_modules/browserify/bin/cmd.js a.js b.js
(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
var b = require('./b.js');
console.log(b.foo());

},{"./b.js":undefined}],2:[function(require,module,exports){
exports.foo = function() {
  return 'works';
};

},{}]},{},[1,2]);
```

Note for just a single file it works:

```
$ node_modules/browserify/bin/cmd.js a.js
[...]

PR #1268, opened 2015-05-16
Fix multi-entry with cross-require

zertosh:
Fixes https://github.com/substack/node-browserify/issues/1266. I broke it in https://github.com/substack/node-browserify/pull/1248. What was happening was that I was setting `row.file` too late, and `row.id` was not getting the full path. The example in  https://github.com/substack/node-browserify/issues/1266 would've worked via the CLI before https://github.com/substack/node-browserify/pull/1248, but not via the API. This PR fixes that.

@jmm I really want to merge the refactor of `b.require` from https://gist.github.com/zertosh/e66a3e96afc45228f02b (taking these changes into account). What tests do you think are missing before we can do it w/o fear of major breakage?

### DIFF
--- a/index.js
+++ b/index.js
@@ -156,16 +156,16 @@ Browserify.prototype.require = function (file, opts) {
     if (typeof file === 'object') {
         row = xtend(file, opts);
     }
-    else if (isExternalModule(file)) {
+    else if (!opts.entry && isExternalModule(file)) {
         // external module or builtin
         row = xtend(opts, { id: expose || file, file: file });
     }
     else {
-        row = xtend(opts, { file: file });
+        row = xtend(opts, { file: path.resolve(basedir, file) });
     }
     
     if (!row.id) {
-        row.id = expose || file;
+        row.id = expose || row.file;
     }
     if (expose || !row.entry) {
         // Make this available to mdeps so that it can assign the value when it
@@ -180,10 +180,7 @@ Browserify.prototype.require = function (file, opts) {
         self._bpack.hasExports = true;
     }
     
-    if (row.entry) {
-        row.file = path.resolve(basedir, row.file);
-        row.order = self._entryOrder ++;
-    }
+    if (row.entry) row.order = self._entryOrder ++;
     
     if (opts.transform === false) row.transform = false;
     self.pipeline.write(row);
--- a/test/multi_entry_cross_require.js
+++ b/test/multi_entry_cross_require.js
@@ -1,17 +1,16 @@
 var browserify = require('../');
 var vm = require('vm');
 var test = require('tap').test;
-var fs = require('fs');
 
 var testFiles = [
-    __dirname + '/multi_entry/a.js',
-    __dirname + '/multi_entry/b.js',
-    __dirname + '/multi_entry/c.js'
+    __dirname + '/multi_entry_cross_require/a.js',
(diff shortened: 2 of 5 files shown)


jmm:
Some that come to mind (don't know which already exist):

* Correct `order` value for streams irrespective of when they finish. What's a good way to simulate a "slow" stream? Can you just do this?:

  ```js
  var stream = through2();
  stream.end("whatever");
  b.require(stream);
  stream.pause();
  b.require('./whatever');
  stream.resume();
  ```

* Correct sham file name (incorporating `order`) for streams.

* Successful exposure of stream (as in `b.require(stream, {expose: 'whatever'})`.

jmm:
Re: the "slow" stream, actually I guess you could just do:

```js
var stream = through2();
stream.write("whatever");
b.require(stream);
b.require('./whatever');
stream.end();
```

zertosh:
Thanks! Those test ideas make sense. I'll write those up before continuing on a refactor, but we should merge this in as a stopgap though.

jmm:
@zertosh 
>  but we should merge this in as a stopgap though.

FWIW I still have the concerns I voiced in https://github.com/substack/node-browserify/pull/1248#issuecomment-99871454, https://github.com/substack/node-browserify/pull/1248#issuecomment-99894906.

zertosh:
@jmm So to summarize your concerns:

* `row.file` needs to be something consistent.
* path resolution when calling `b.require` needs to be clear (whether it's fs-like or require-like, doesn't matter - just clear).

We just need decisions, right? Because it's not a technical problem - it's easy to code up. How do we decide? Is it our place to decide? And are you ok if I merge this to fix the immediate problem?

jmm:
@zertosh 

* `row.file` needs to have well-defined, and ideally consistent, semantics and type. Same for `row.id`.

* path resolution semantics for "root" files (the ones that the dependency graph starts from -- `files`, `opts.entries`, `b.add()`, `opts.require`, `b.require()`) need to be clear, including documentation and tests. Even #1248 illustrates how easy it is to get tripped up currently, as it refers to resolving values to absolute paths, which wasn't actually the case before or after (and has important implications for consistency elsewhere).

* In my opinion decisions about the API should be made with little regard for the CLI. API consistency with the CLI is not an imperative. A well-designed API should be the imperative. Ultimately the CLI is just a consumer of the API that needs to special-case and pre-process its input to massage it into the appropriate format for passing to API calls. Sometimes there's enough overlap between what the CLI and API are doing that the code can just live in the API implementation and doesn't need to be handled specially by the CLI and that's a bonus.

> We just need decisions, right?

Right.

> How do we decide? Is it our place to decide?

I think it's fruitful for collaborators or any interested parties to discuss it, but if it's not clear cut what should happen I suppose @substack would need to make the call.

> And are you ok if I merge this to fix the immediate problem?

Ultimately it's your decision as you don't need my approval. Until these issues get legitimately dealt with I think it would make more sense to revert to the previous situation, where the CLI pseudo-resolved the paths. It's important to keep in mind that #1248 *changed* the resolution behavior of the API. Maybe the previous resolution behavior was an accident, maybe not, but I think we should try to minimize churn until those questions are deliberately decided. Right now we're chasing our tails with these kinds of changes. It wouldn't surprise me if these changes, while fixing one problem, cause other unexpected problems, even beyond the already noted change to resolution behavior.

PR #1269, opened 2015-05-16
setup globarTr once per bundle and only if needed

zertosh:
This PR optimizes the globals transform by:

  1. Only adding it to the pipeline if it hasn't been opted-out of. This check was previously in the transform itself.
  2. Only setting up the options for `insertGlobals` once per `bundle()`, instead of once for every file.

cc: @jmm

### DIFF
--- a/index.js
+++ b/index.js
@@ -510,43 +510,7 @@ Browserify.prototype._createDeps = function (opts) {
     });
     
     mopts.globalTransform = [];
-    if (!this._bundled) {
-        this.once('bundle', function () {
-            self.pipeline.write({
-                transform: globalTr,
-                global: true,
-                options: {}
-            });
-        });
-    }
-    
-    var no = [].concat(opts.noParse).filter(Boolean);
-    var absno = no.filter(function(x) {
-        return typeof x === 'string';
-    }).map(function (x) {
-        return path.resolve(basedir, x);
-    });
-    
-    function globalTr (file) {
-        if (opts.detectGlobals === false) return through();
-        
-        if (opts.noParse === true) return through();
-        if (no.indexOf(file) >= 0) return through();
-        if (absno.indexOf(file) >= 0) return through();
-        
-        var parts = file.split('/node_modules/');
-        for (var i = 0; i < no.length; i++) {
-            if (typeof no[i] === 'function' && no[i](file)) {
-                return through();
-            }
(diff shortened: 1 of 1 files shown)


jmm:
So there's no legit case for opts changing between `globalTr()` calls right? If so, LGTM. A few things that I noticed while looking at this:

* If there's going to be a new method (`_createGlobalTr`) I'd be tempted to put all of the logic in that method. In other words move [this part](https://github.com/substack/node-browserify/blob/28420c28ef676d25d0a696e44b9c0cb1c16b3352/index.js#L513-L521) into it.

* [`debug: opts.debug`](https://github.com/substack/node-browserify/pull/1269/files#diff-168726dbe96b3ce427e7fedce31bb0bcR536), seems superfluous? (I know it was already like this.)

* [Assigning `mopts.globalTransform = [];`](https://github.com/substack/node-browserify/pull/1269/files#diff-168726dbe96b3ce427e7fedce31bb0bcR512) seems superfluous? (I know it was already like this.)

* module-deps `opts.globalTransform` should be documented.

zertosh:
> So there's no legit case for opts changing between globalTr() calls right?

You know, this is weird. I've always considered a browserify instance "sealed" once you call `bundle` for the first time. <del>Technically, you can always add more requires, excludes, ignores, transforms, etc.</del> (the entry order would get messed up). However, even before this change, `opts` was always the same copy `globalTr` had in scope when it was first created - so even if you added more opts, it never saw them. But if you mutated `noParse`, it would see that change since that array would be available by reference. But yes, no legit case.

> If there's going to be a new method (_createGlobalTr) I'd be tempted to put all of the logic in that method. In other words move this part into it.

I undid that. It doesn't follow the general code structure. There is not reason for `_createGlobalTr` to be a method on `b` since it doesn't need `this`. If I move the logic you suggested, then that method wouldn't return anything useful, and that would be a first in the code. Public methods return `this` so you can chain (minus `bundle` and `reset`), and the private methods return streams (most, if not all).

> debug: opts.debug, seems superfluous?

good catch!

> Assigning mopts.globalTransform = []; seems superfluous? 

I think that's so you don't use `opts.globalTransform` to pass transforms to module-deps. That's something to figure out later.

> module-deps opts.globalTransform should be documented.

It should die instead! heheh too many options.

zertosh:
I've got another idea :smile:

PR #1273, opened 2015-05-20
Housekeeping

zertosh:
Just a few little things that have fallen through the cracks.


zertosh:
I removed https://github.com/substack/node-browserify/commit/5107d0b7fcbd24009bacfed77560e735ce4a8a0c because it turns out that in node 0.8 `process.env` does not have `hasOwnProperty`. TIL

Issue #1290, opened 2015-06-01
Allow global variables to be declared without including `process`


sedge:
[From the looks of it](https://github.com/substack/node-browserify/blob/master/index.js#L548-L550) there's no way to specify which globals to include without including `process` and therefore a `process` shim. Why is this? Am I missing something?

hawkerboy7:
I think I have this same problem in my issue: https://github.com/substack/node-browserify/issues/1277. Changing options I either get `Uncaught Error: Cannot find module '_process' ` or `fs == {} and then process does work`

sedge:
I got around this by adding an undefined `process` definition to `opts.insertGlobalVars`:

```js
var options = {
  insertGlobalVars: {
    process: function() { return; }
  }
};
```

hawkerboy7:
Thanks this worked!!

t128:
thank you @sedge that fixed my build #1277 :+1:

Issue #1287, opened 2015-05-29
--insert-globals doesn't work?


cachaito:
Hi I tried to attach some global variables through npm script in that way:
```
"scripts": {
    "prewatch:js": "browserify app/js/app.js -o out/bundle.js -d --insert-global-vars PORT=9999"
}
```
but in browser `process.argv` or `env` properities are empty :confused:

axkibe:
I just an into this issue as well. --igv does not work.

One can see in bin/args.js why:
```
    if (argv.igv) {
        var insertGlobalVars = {};
        var wantedGlobalVars = argv.igv.split(',');
        Object.keys(insertGlobals.vars).forEach(function (x) {
            if (wantedGlobalVars.indexOf(x) === -1) {
                insertGlobalVars[x] = undefined;
            }
        });
    }
```

There is no way a "wantedGlobalVar" is going to get into insertGlobalVars, the only thing it does is disable those default globals that are not in --igv, but a new global that is not a default global is never going into it.

PS: My usecase is different I got a global "DEBUG" variable. Removing DEBUG code for the release works well with:
```browserify src/opentype.js -p licensify --bare --standalone opentype -g [ uglifyify --compress [ --dead_code --global_defs [ --DEBUG 0 ] ] ] > dist/opentype.min.js```

But for the not minified version, DEBUG is not defined and since I'm using 'strict mode' it raises an exception. :-/

Issue #979, opened 2014-11-07
No globals with detectGlobals, insertGlobals config


JosephClay:
I am bundling a project using:

```
browserify('path/to/file', { insertGlobals: true, detectGlobals: false });
```

The intent is to have a faster build, and have globals inserted. However, the build is faster but no globals (`process,global,Buffer,__argument0,__argument1,__argument2,__argument3,__filename,__dirname`) are inserted.

goto-bus-stop:
Setting `detectGlobals` to false disables inserting globals entirely; just `browserify('/path/to/file', { insertGlobals: true })` does the trick.

PR #1298, opened 2015-06-06
define __dirname,__filename as documented for --bare, ...

arhart:
... but don't define process,global,Buffer.

This fixes substack/covert#12

This might also be of interest to:
* substack/node-browserify#1287
* substack/node-browserify#1290
* substack/node-browserify#979

### DIFF
--- a/bin/args.js
+++ b/bin/args.js
@@ -64,6 +64,14 @@ module.exports = function (args, opts) {
         return entry;
     });
     
+    var insertGlobalVars;
+    if (argv.igv) {
+        insertGlobalVars = argv.igv.split(',').reduce(function (vars, x) {
+            vars[x] = insertGlobals.vars[x];
+            return vars;
+        }, {});
+    }
+    
     if (argv.node) {
         argv.bare = true;
         argv.browserField = false;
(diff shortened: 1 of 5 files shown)


AaronO:
@arhart Great work but this still has a bug, it always inserts `process` as a global variable.

Run this test to see for yourself:

```
echo "var a = 88;" | browserify --no-bundle-external --ig --igv "__dirname" -
```

Any idea what causes `process` to be added even when not specified ?

arhart:
@AaronO The old code called out process and buffer (not Buffer) to get default handling from "https://github.com/substack/insert-module-globals/". The changes in acbb2f7 ignore buffer and call out only process for non-insertion. I'm not sure this is correct either, but it seems closer to what you're looking for.

More tests around what --no-bundle-external should/shouldn't do might be good. Maybe that's a subject for another issue and pull-request?

zertosh:
hey @arhart, sorry, I'm having a hard time understand why so many changes are needed for this (maybe bc https://github.com/substack/insert-module-globals/issues/50 wasn't merged?). Is https://github.com/substack/node-browserify/commit/01bc208cc9bc5ea3fe618a3b5786999f0890bd86 close to what you're fixing?

arhart:
@zertosh merging substack/insert-module-globals#50 simplified some. Changing bare to not set argv.detectGlobals = false simplified more. In any case, 01bc208 looks good, thanks. I've rebased it and pushed it to my branch in this pull request.

zertosh:
@arhart Looks good! Sorry about the delay - totally missed this

zertosh:
So this is all good for `--bare`, but I just realized that `--insert-global-vars` is really broken :( I'm going to look into that before merging this

PR #1317, opened 2015-06-30
add dedupeInstance option

### LINKED ISSUE #1063: Modules are executed multiple times when using symlinks in node_modules
I have following symlink in my project

    ln -sf .. node_modules/app

and I require modules like this

```js
var config = require("app/config");
```
etc.

but in this case the config module is executed multiple times when it is required from different js files.

Here's smallest possible example reproducing this https://github.com/epeli/browserify-bug

This bug was introduced in the 8.0.0 release.

--

The symlink idea is from the browserify handbook https://github.com/substack/browserify-handbook#symlink

chrisirhc:
Resolves #1063 if user opts in to use `dedupeInstance`.

Note that there are no new tests, this is to request for further comment. If community approves, I'll add tests. The tests were there in pre-8.0.0 anyway.

### DIFF
--- a/index.js
+++ b/index.js
@@ -384,7 +384,7 @@ Browserify.prototype._createPipeline = function (opts) {
         'unshebang', [ this._unshebang() ],
         'syntax', [ this._syntax() ],
         'sort', [ depsSort(dopts) ],
-        'dedupe', [ this._dedupe() ],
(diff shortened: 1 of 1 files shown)


chrisirhc:
Closing as #1318 is a better solution

PR #1318, opened 2015-06-30
use same instance on symlinked modules

### LINKED ISSUE #1063: Modules are executed multiple times when using symlinks in node_modules

chrisirhc:
Handle required packages by their real path.

Resolves #1063

Again, this is a request for further comment.

This isn't using a cache which might be faster (or not, since resolve might only be called for new packages encountered).


chrisirhc:
Updated PR. Tests now pass.

substack:
Aside from the sync issue, this looks good.

substack:
I added some tests for the case in #831 and everything works. Published in 10.2.5/10.2.6. Thanks for the patch!

marcello3d:
This update is breaking our build. :-( 

Problem: 

1. We have two modules, `app` and `shared`
2. `shared` is symlinked into `app`
3. Both `app` and `shared` use a module `foo`, which is installed in `app/node_modules/foo` (but not `shared/node_modules`)
4. Now `shared` can't find `foo`

PR #1321, opened 2015-07-01
Update readme.markdown

mjhm:
Change to non deprecated link.


terinjokes:
Thanks for catching that.

PR #1362, opened 2015-08-24
bump buffer to 3.4.3 to fix maximum call stack exceeded errors

mnichols:
Many modules are reporting `maximum call stack exceeded` errors (at least in Chrome) that were due to a bug in the `buffer` module.
eg:
- https://github.com/evanw/node-source-map-support/issues/95
- https://github.com/feross/buffer/pull/70

This PR simply bumps buffer to the 3.4.3 revision and has fixed the source maps problem as well as another problem I came across in auth0's lock module.

All tests pass with this bump

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "browser-pack": "^5.0.0",
     "browser-resolve": "^1.7.1",
     "browserify-zlib": "~0.1.2",
-    "buffer": "^3.0.0",
+    "buffer": "^3.4.3",
     "builtins": "~0.0.3",
     "commondir": "0.0.1",
     "concat-stream": "~1.4.1",


substack:
The semver: `"buffer": "^3.0.0"` means that npm will install 3.4.3 on a clean install.

mnichols:
eek I forget about doing an `npm cache clean` before patching. Still, does it make sense to allow lower versions of buffer that have such a insidious bug in it to be brought in? For example, when using node-source-map support this [bug](https://github.com/evanw/node-source-map-support/issues/95) bit me and it was tricky to track down.

feross:
I'm fine with preventing lower buffer versions from being used to prevent the bug in buffer 3.4.2.

PR #1365, opened 2015-08-25
Add failing test for top level transforms and symlinked node modules

ivantsepp:
Adds a failing test with using a transform on top-level files on symlinked node modules. 

The test failure is coming from [passing in the realpath](https://github.com/substack/node-browserify/blob/f7e626ca75bd925a584f9b0fe8ff9d4b8e90f188/index.js#L486) and `module-deps` uses this path to [determine if the file is top level](https://github.com/substack/module-deps/blob/9112e326d76a97869ecffee06d84be21e6f7d157/index.js#L200-L216).

A possible solution is to not pass in the realpath and calculate it instead in `module-deps`. I think this makes sense as `module-deps` would want to know the realpath **and** the resolved path of the file.


ivantsepp:
Any updates on this? Is the test in this PR expected behavior?

jmm:
@ivantsepp Thanks for this! I haven't tried your test (and unfortunately I can't manage to fully "get it" just from the description), but there's some discussion going on in #1386 that you may be interested in.

zertosh:
I guess this is fixed by https://github.com/substack/node-browserify/pull/1392

chrisirhc:
@zertosh Yep, though #1392 hasn't been merged.

zertosh:
@chrisirhc cool! Lets let ppl chime in on #1413 since that'll bump browserify to v12, so then #1392 can ride along with that.

PR #1377, opened 2015-09-09
Update travis Node.js version & remove io.js

demohi:


zertosh:
Thanks!

terinjokes:
I can open a new issue, but should we consider using `- 4` instead of `- "4.0.0"` to allow nvm to use the latest stable release of the 4.x series of node?

zertosh:
ah yeah... good call

Issue #1137, opened 2015-02-23
Is exposed name from `b.require()` supposed to be available inside bundle?


jmm:
The documentation for `b.require()` says:

> Make file available from outside the bundle with require(file).

This doesn't communicate any relationship between `b.require()`d files and other files in the same bundle. However, people frequently use `b.require()` to expose a file under an alias that they then use to require the file in other files in the bundle, e.g.:

`bundler.js`
```js
browserify('./entry').require('./x', {expose: "whatever"});
```

`entry.js`
```js
require('whatever');
```

I believe some of browserify's tests of `b.require()` even rely on that.

If that behavior is intended it should be documented, and if it's not intended the documentation should note that to make it clear. I think it's a useful feature, if it doesn't interfere with any other use cases. There is currently a lot of flakiness in browserify about matching up values passed / configured via things like `b.require()`, `require()` calls in bundled files, `browser` field etc. though, so getting the documentation tightened up re: what should happen would be very beneficial.

PR #1391, opened 2015-09-23
Only use entry files for standalone export

### LINKED ISSUE #1120: .add or .require expose option hijacks entry module
If you `b.add` or `b.require` a module and re-map it's module id using the `expose` option, it will replace the `entry` module as the root module for the bundle. I reproduced this in a demonstration repo: https://github.com/asakusuma/browserify-expose-bug

I'm assuming this is not expected behavior, but if it is, is there another way to include a file as a module and re-map the module id?

### LINKED ISSUE #1390: standalone - highest number always wins
If I have two entry files, then the one with the highest module number always ends up being the standalone file. Can we make this deterministic instead? First file added? The ones added by `add()` and not `require()`? Just something so I can control it.

ptarjan:
If I want to to use `require('foo')` in my code, I do `browserify().require('./foo', {expose: 'foo'})`. Just because I do that, doesn't mean I want that to now be an entrypoint of my app, I just want it to be a library available internally.

Fixes #1390
Fixes #1120

This breaks backwards compatibility for people solely using `require()` (which I think is an error since there is not entry files). Change the one that was highest number to an `add()` and it will work as before.

### DIFF
--- a/index.js
+++ b/index.js
@@ -712,7 +712,7 @@ Browserify.prototype._label = function (opts) {
             }
         });
         
-        if (row.entry || row.expose) {
+        if (row.entry) {
             self._bpack.standaloneModule = row.id;
         }
         this.push(row);
--- a/test/debug_standalone.js
+++ b/test/debug_standalone.js
@@ -41,15 +41,11 @@ test('debug standalone', function (t) {
     });
 });
 
-test('debug standalone exposed', function (t) {
+test('debug standalone entry', function (t) {
     t.plan(2);
-    
-    var stream = through();
-    stream.push('console.log(1+2)');
-    stream.push(null);
-    
+
     var b = browserify({ debug: true, standalone: 'xyz' });
-    b.require(__dirname + '/debug_standalone/x.js', { expose: 'xxx' });
+    b.add(__dirname + '/debug_standalone/x.js');
     b.bundle(function (err, buf) {
         var src = buf.toString('utf8');
         var last = src.split('\n').slice(-2)[0];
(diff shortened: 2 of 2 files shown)


jmm:
@ptarjan 
> This breaks backwards compatibility for people solely using `require()` (which I think is an error since there is not entry files). Change the one that was highest number to an `add()` and it will work as before.

I'm not sure it's that straightforward. `b.require()` and `b.add()` are somewhat use interchangeably in ways that aren't actually documented (e.g. by passing opts that blur the lines between them). And the [handbook says](https://github.com/substack/browserify-handbook#standalone):

> Note however that standalone only works with a single entry or directly-required file.

I interpret "directly-required" to mean `b.require()`'d.

Also, your usage of `b.require()` ("I just want it to be a library available internally"), while common, is [undocumented](1137).

I think there needs to be more clarity around these issues before it can be fixed properly. I think even the term "entry file" is maybe not a great choice. Does it mean entry to a dependency graph (e.g. `b.require('foo')`) or a script that runs when the bundle loads?

ptarjan:
@jmm well if it is really that fast and loose with `require()` and `add()` and we don't want to break backwards compat with how people are using it, how would you feel about a new function `.standalone(fileName, exportName)` which in incompatible with the `browserify({standalone: true})` and basically calls `.add(fileName)` and forces it to be the exported file?

jmm:
@ptarjan That's one of the ideas I floated in https://github.com/substack/node-browserify/issues/1120#issuecomment-74084569, but I can't unilaterally change or expand the API. And in any case, for my own consideration, I'd want more information and better documentation about the existing API (e.g. #1137) before coming to conclusions about how to handle this issue.

ptarjan:
@jmm ok, well I'll leave it in your capable hands and hack around it somehow

PR #1392, opened 2015-09-24
do not apply root transforms on linked modules

### LINKED ISSUE #1386: browserify applies transforms to npm linked modules
If you `npm link` a module into your browserify project, it incorrectly has transforms applied to it. This becomes a major issue particularly if the module you install relies on non-strict mode and you're using babelify, which inserts `"use strict"`, causing it to error.

Here's a repo to reproduce: https://github.com/jayphelps/browserify-bug. The major step to note in the README is that you must git clone the external module and `npm link` it for the bug to show.

Also note that very new versions of babel proactively transform certain strict mode code, like `this.foo` to `undefined.foo`. Don't let this confuse you.

chrisirhc:
Incudes tests from #1365 and fixes #1365 , fixes #1386.

This *will* fail until https://github.com/substack/module-deps/pull/99 is merged and released on module-deps. I'll re-run when that's done.


chrisirhc:
Updated to use module-deps 4.0.0, so that tests pass.

substack:
Merged into 12.0.1. I thought we had this in 12.0.0 but I think it slipped through the cracks. Thanks for all the great work!

ivantsepp:
:tada: thanks for merging and thanks @chrisirhc for the fix :smile:

PR #1393, opened 2015-09-26
Links to the wrong querystring module

emilbayes:
Previously linked to `querystring` while it actually depends on `querystring-es3`


feross:
thanks!

PR #1401, opened 2015-10-11
Bumped dep to commondir@1.0.1

cbas:
Fixes warning about non-SPDX license usage in commondir@0.0.1


zertosh:
This dep isn't used. Fixed by https://github.com/substack/node-browserify/pull/1413

PR #1402, opened 2015-10-12
Bumping labeled-stream-splicer version

cbas:
Eliminates dependency on obsolete `indexof@0.0.1` (deprecated by `component-indexof`) in outdated sub-depedency `stream-splicer@1.3.2`.


zertosh:
Fixed by https://github.com/substack/node-browserify/pull/1413

zertosh:
Thanks for these PRs though. The reason they have to be merged in #1413 is because they're part of the streams3 upgrade.

PR #1403, opened 2015-10-12
Version bump

cbas:
The outdated `stream-http@1.7.1` has a dependency on the  deprecated
`indexof@0.0.1` package. See also: https://github.com/substack/node-browserify/pull/1402


zertosh:
Fixed by https://github.com/substack/node-browserify/pull/1413

PR #1405, opened 2015-10-16
Remove unnecessary workaround to hide full path

parshap:
This was previously added in 5fe121b to prevent the absolute path to the "process" module from exposed in the bundle.

Recently a [change landed in insert-module-globals][1] that prevents this from happening in the first place, thus this workaround is no longer needed.

[1]: https://github.com/substack/insert-module-globals/pull/54

This is technically a breaking change for those who are calling `require("_process")` directly (which should not happen as it was never advertised as a public api).

**There are currently failing tests** 

```
    not ok 2 - cwd directory visible
      ---
      found: 2675
      wanted: -1
      compare: ===
      at:
        file: test/leak.js
        line: 32
        column: 11
      source: |
        t.equal(src.indexOf(process.cwd()), -1, 'cwd directory visible');
      stack: >
        test/leak.js:32:11
        ...
```

This is because the test checks for the existence of process.cwd()` in the bundle, which exists because the basedir of the bundle is set to `/tmp/...` and the relative path from `/tmp/...` to where the process module is happens to include the `process.cwd()` substring in it: `../../../home/parshap/projects/node-browserify/node_modules/process/browser.js`.

I'm not sure what the right approach is here.

1. Is this ok, or we are allowing too much to be exposed in the odd case that the bundle basedir is in a completely different directory than where the browserify `node_modules` directory is?

1. If it's ok, how should the test be changed? Should this assertion be removed?

1. If it's not ok, should we add the same kind of workaround for `Buffer.isBuffer` (e.g., ass a `require("_is-buffer")` builtin)?

### DIFF
--- a/index.js
+++ b/index.js
@@ -547,9 +547,7 @@ Browserify.prototype._createDeps = function (opts) {
             }
         }
         
-        var vars = xtend({
-            process: function () { return "require('_process')" },
-        }, opts.insertGlobalVars);
+        var vars = xtend(opts.insertGlobalVars);
         
         if (opts.bundleExternal === false) {
             vars.process = undefined;
--- a/lib/builtins.js
+++ b/lib/builtins.js
@@ -35,4 +35,3 @@ exports.url = require.resolve('url/');
 exports.util = require.resolve('util/util.js');
 exports.vm = require.resolve('vm-browserify');
 exports.zlib = require.resolve('browserify-zlib');
-exports._process = require.resolve('process/browser');


parshap:
This was previously discussed in https://github.com/substack/insert-module-globals/pull/31 and https://github.com/substack/insert-module-globals/pull/33.

parshap:
Have thought about this a bit and my current thoughts are:

1. Keep using relative paths in *insert-module-globals*. I can't think of advantages in using absolute paths, and relative paths will be *at least* as good as absolute paths and will usually hide more information about the system than absolute paths.

1. Keep the the `_process` builtin and add a `_is-buffer` one as well. This is an abstraction leak, but will guarantee no system paths are exposed.

Longterm, it'd be nice if *insert-module-globals* handled this by itself. Maybe it should become a plugin instead of a transform so that it can inject builtins itself. (Can plugins do this? I think so.) Or maybe there's a better approach.

zertosh:
As far as using relative paths for `process` and `buffer` (with the updated insert-module-globals), I think it's great. With relative paths you get deterministic builds in projects that have a local browserify (most everyone?).

As far as the information leaking that using `_process` aims to protect from: Before this change, you'd leak info if you used a local browserify, but not any info about your project if you used a global browserify. This change reverses that. Having `basedir` be outside of the project directory is like using a globally installed browserify.

I'm totally in favor of this change since projects tend to use their own browserify. The test should be modified so that `basedir` is the `cwd` (so just omit it), and that's it.

parshap:
> As far as using relative paths for process and buffer (with the updated insert-module-globals), I think it's great. With relative paths you get deterministic builds in projects that have a local browserify (most everyone?).

Indeed. Non-deterministic builds was what led to me to discover this issue in the first place.

> As far as the information leaking that using _process aims to protect from: Before this change, you'd leak info if you used a local browserify, but not any info about your project if you used a global browserify. This change reverses that. Having basedir be outside of the project directory is like using a globally installed browserify.

Well, currently you don't leak *any* information because of the `_process` hack. (But such a thing does not exist for `is-buffer`). With the *insert-module-globals* change, even without `_process`, you still wouldn't expose anything, other than relative paths within your project, as long as you're using a local browserify. I'm a little worried about those using a global browserify though and I'd wager it's a non-trivial amount of users.

parshap:
> I'm totally in favor of this change since projects tend to use their own browserify. The test should be modified so that basedir is the cwd (so just omit it), and that's it.

I've made this change to the tests. This is ready to merge as long as we're okay with the implications to those using a global browserify (or doing some weird `NODE_PATH` stuff).

PR #1413, opened 2015-10-26
Mostly streams3 updates (should be v12)

zertosh:
The commits are pretty self-explanatory.

For the changelog:

```
# 12.0.0

Node support changes: now testing against the latest node (currently 4.2.1). node 0.8 is no longer supported, and iojs isn't actively tested.

Stream3 everywhere! Everything has been updated to use streams3.

Sourcemaps charset now uses an `=` instead of a `:`. This fixes certain issues with Chinese characters in sourcemaps. See #753.

module-deps has been updated to fix root transforms on symlinked modules. See https://github.com/substack/module-deps/pull/99.

stream-http, the module that provides `http` support, dropped IE8 support. If you depend on this, see https://github.com/jhiesey/stream-http#ie8-note.

Removed `builtins` and `commondir` – both unused dependencies.
```


zertosh:
Also merge in: https://github.com/substack/node-browserify/pull/1392

package.json
```diff
@@ -22,63 +22,61 @@
...
-    "duplexer2": "~0.0.2",
-    "events": "~1.0.0",
-    "glob": "^4.0.5",
+    "duplexer2": "~0.1.2",
+    "events": "~1.1.0",
+    "glob": "^5.0.15",
     "has": "^1.0.0",
     "htmlescape": "^1.1.0",
-    "stream-http": "^1.2.0",
+    "stream-http": "~2.0.0",
```
feross: Can we change this to `^`? @jhiesey's following semver.

zertosh: :+1:

feross:
LGTM if all the tests pass

substack:
Merged, thanks!

Issue #1217, opened 2015-04-16
gulp + vinyl-transform + browserify - broken, last working 9.0.4


Kruegs:
Hello!

I've been getting the below error since upgrading beyond 9.0.4.  I've also included my code for the gulp task further below.  It works with version 9.0.4.

I appreciate any help,
John

```
[17:16:46] Starting 'browserify'...
_stream_readable.js:540
    var ret = dest.write(chunk);
                   ^
TypeError: undefined is not a function
    at Producer.ondata (_stream_readable.js:540:20)
    at Producer.emit (events.js:107:17)
    at Producer.Readable.read (_stream_readable.js:373:10)
    at flow (_stream_readable.js:750:26)
    at resume_ (_stream_readable.js:730:3)
    at _stream_readable.js:717:7
    at process._tickCallback (node.js:355:11)
```

```
gulp.task("browserify", function() {
    var browserify     = require("browserify");
    var vinylTransform = require("vinyl-transform");

    var buildBundle = function () {
        return vinylTransform(function (filename) {
            var b = browserify(filename);
            return b.bundle();
        });
    };

    return gulp.src("src/js/*.js")
        .pipe(buildBundle())
        .pipe(gulp.dest("target/js"));
});
```

terinjokes:
This was an error in the gulp recipes that relied on internal behavior. (See #1191 and #1044). Please see the updated gulp guides:

* [Browserify + Uglify2](https://github.com/gulpjs/gulp/blob/4d35560d9e2e992037886897c671518cfe49fd7f/docs/recipes/browserify-uglify-sourcemap.md#browserify--uglify2-with-sourcemaps)
* [Browserify + Globs](https://github.com/gulpjs/gulp/blob/4d35560d9e2e992037886897c671518cfe49fd7f/docs/recipes/browserify-with-globs.md#browserify--globs)

Issue #1213, opened 2015-04-13
version >= 9.0.3 fails with gulp and vinyl-transform


ibc:
Node 0.12.0.

```
    "browserify": "^9.0.8",
    "gulp": "git+https://github.com/gulpjs/gulp.git#4.0",
    "vinyl-transform": "^1.0.0"
```

gulpfile.js:

```
javascript
gulp.task('browserify', function() {
	var browserified = vinyl_transform(function(filename) {
		var b = browserify(filename, {
			standalone: pkg.name
		});
		return b.bundle();
	});

	var src = pkg.main;
	return gulp.src(src)
		.pipe(browserified)
		.pipe(header(banner, banner_options))
		.pipe(rename(builds.uncompressed))
		.pipe(gulp.dest('dist/'));
});
```

This just worked with previous versions of browserify. But now I get:

```
[15:58:42] 'browserify' errored after 68 ms
[15:58:42] TypeError: undefined is not a function
    at Producer.ondata (_stream_readable.js:540:20)
    at Producer.emit (events.js:107:17)
    at Producer.Readable.read (_stream_readable.js:373:10)
    at flow (_stream_readable.js:750:26)
    at resume_ (_stream_readable.js:730:3)
    at _stream_readable.js:717:7
    at process._tickDomainCallback (node.js:381:11)
```

terinjokes:
This was an error in the gulp recipes that relied on internal behavior. (See #1191 and #1044). Please see the updated gulp guides:

* [Browserify + Uglify2](https://github.com/gulpjs/gulp/blob/4d35560d9e2e992037886897c671518cfe49fd7f/docs/recipes/browserify-uglify-sourcemap.md#browserify--uglify2-with-sourcemaps)
* [Browserify + Globs](https://github.com/gulpjs/gulp/blob/4d35560d9e2e992037886897c671518cfe49fd7f/docs/recipes/browserify-with-globs.md#browserify--globs)

ibc:
Thanks a lot.

Issue #1199, opened 2015-04-06
Error in 9.0.6+ seems related to readonly streams


torarvid:
Hey. We use browserify via gulp to build a libs.js containing all our 3rd party deps. This works fine in 9.0.4, but in 9.0.6 and 9.0.7 I get:

    _stream_readable.js:524
        var ret = dest.write(chunk);
                       ^
    TypeError: undefined is not a function
        at Producer.ondata (_stream_readable.js:524:20)
        at emitOne (events.js:77:13)
        at Producer.emit (events.js:166:7)
        at Producer.Readable.read (_stream_readable.js:356:10)
        at flow (_stream_readable.js:734:26)
        at resume_ (_stream_readable.js:714:3)
        at _stream_readable.js:701:7
        at process._tickDomainCallback (node.js:366:13)

terinjokes:
This was an error in the gulp recipes that relied on internal behavior. (See #1191 and #1044). Please see the updated gulp guides:

* [Browserify + Uglify2](https://github.com/gulpjs/gulp/blob/4d35560d9e2e992037886897c671518cfe49fd7f/docs/recipes/browserify-uglify-sourcemap.md#browserify--uglify2-with-sourcemaps)
* [Browserify + Globs](https://github.com/gulpjs/gulp/blob/4d35560d9e2e992037886897c671518cfe49fd7f/docs/recipes/browserify-with-globs.md#browserify--globs)

You may also wish to consider using the Browserify API directly, and forgoing gulp entirely.

spicydonuts:
> You may also wish to consider using the Browserify API directly, and forgoing gulp entirely.

Why might one want to consider doing that?

terinjokes:
> Why might one want to consider doing that?

Keeping things simple, and not having to deal with integrating with gulp.

spicydonuts:
Less work for sure, but I don't think people are setting up gulp for fun.

Issue #1191, opened 2015-04-03
9.0.6 breaks my gulp + watchify setup


msutherl:
```
@Gulpfile.js

var gulp       = require('gulp');
var gutil      = require('gulp-util');
var transform  = require('vinyl-transform');
var debug      = require('gulp-debug');
var browserify = require('browserify');
var watchify   = require('watchify');
var reactify   = require('reactify');
var _          = require('underscore');

var paths = {
  src:  './www/src/',
  dist: './www/dist/'
};

var b = browserify(paths.src + 'js/', _.extend(watchify.args, {debug: true}));
var bundler = watchify(b)
  .transform(reactify)
  .on('update', bundle)
  .on('log', console.error);

console.log(watchify.args);

function bundle() {
  var bundle = transform(function(filename) {
    return bundler.bundle();
  });
  return gulp.src([paths.src + 'js/index.js'])
    .pipe(debug({title: 'bundle'}))
    .pipe(bundle)
    .on('error', gutil.log.bind(gutil, 'Browserify Error'))
    .pipe(gulp.dest(paths.dist));
}

gulp.task('browserify', bundle);
```

Upgrading to 9.0.6 results in the following error:

```
_stream_readable.js:540
    var ret = dest.write(chunk);
                   ^
TypeError: undefined is not a function
    at Producer.ondata (_stream_readable.js:540:20)
    at Producer.emit (events.js:107:17)
    at Producer.Readable.read (_stream_readable.js:373:10)
    at flow (_stream_readable.js:750:26)
    at resume_ (_stream_readable.js:730:3)
    at _stream_readable.js:717:7
    at process._tickCallback (node.js:355:11)
```

9.0.4 works fine

valette:
+1 here : 
```
var browserify = require('browserify'),
    gulp       = require('gulp'),
    transform  = require('vinyl-transform'),
    uglify     = require('gulp-uglify');
    
gulp.task('default', function () {
	var browserified = transform(
		function(filename) {
			return browserify(filename).bundle();
		});

	return gulp.src([__dirname + '/lib/browserified.js'])
		.pipe(browserified)
		.pipe(uglify())
		.pipe(gulp.dest(__dirname + '/cache'));
});
```

terinjokes:
[...]

PR #1415, opened 2015-10-26
Dep round trip

thedufer:
This fixes https://github.com/jsdf/browserify-incremental/issues/22, which arises because `dep` events come with a row that can't be fed back into the cache.  This happens in the case when you have a file that is both exposed and required under its exposed name in the same bundle (which seems like it ought to be valid, and works fine without setting the cache option).

### DIFF
--- a/index.js
+++ b/index.js
@@ -448,6 +448,7 @@ Browserify.prototype._createDeps = function (opts) {
         if (opts.filter && !opts.filter(id)) return false;
         if (self._external.indexOf(id) >= 0) return false;
         if (self._exclude.indexOf(id) >= 0) return false;
+        if (self._expose.hasOwnProperty(id)) return false;
         if (opts.bundleExternal === false && isExternalModule(id)) {
             return false;
         }
--- a/test/dep.js
+++ b/test/dep.js
@@ -23,3 +23,23 @@ test('dependency events', function (t) {
         return a.id < b.id ? -1 : 1;
     }
 });
+
+test('round-trip dependency events', function(t) {
+    var b = browserify(__dirname + '/entry/needs_three.js');
+    var cache = {};
+    b.on('dep', function(row) {
+        cache[row.file] = row;
+    });
+
+    b.require(__dirname + '/entry/three.js', { expose: 'three' });
+    b.bundle(function(err, src) {
+        var b2 = browserify(__dirname + '/entry/needs_three.js', { cache: cache });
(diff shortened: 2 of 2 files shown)


jmm:
@thedufer Thanks for contributing to browserify! I haven't looked at your PR enough to comment on anything else, but re:

> This happens in the case when you have a file that is both exposed and required under its exposed name in the same bundle (which seems like it ought to be valid

That actually is [not documented](https://github.com/substack/node-browserify/issues/1137). In my opinion nothing related to that should be changed without it first being documented. I think it's intended to work that way, and therefore should be documented, but I've gotten no feedback on my issue.

thedufer:
We definitely depend on that behavior, although I do now remember that we had to do some testing to make sure it did what we hoped, because of the lack of documentation.  I agree that it should be documented.  Anything I can do to help?  What kind of feedback would you want on that issue before its worth fixing up the documentation?

jmm:
@thedufer 
> Anything I can do to help?

Probably not, unfortunately, but thank you for asking. :smiley:

> What kind of feedback would you want on that issue before its worth fixing up the documentation?

It's definitely worth fixing (if there's something to be fixed). I'm looking for feedback from @substack or at least a couple of other Collaborators. I can make a guess, even an educated one, on my own about how it's *intended* to work, but that's a haphazard basis for changing API documentation (and that can be tantamount to changing the API itself, which as far as I know no Collaborator should be doing unilaterally).

> We definitely depend on that behavior, although I do now remember that we had to do some testing to make sure it did what we hoped, because of the lack of documentation.

I'd give all the usual cautions about relying on undocumented behavior. People constantly want to do things that are undocumented and that do actually work the way they'd like, until it breaks in a future release (even a patch). See for example (they're all about the same undocumented behavior that lots of people were relying on):

https://github.com/substack/node-browserify/issues/1044
https://github.com/substack/node-browserify/issues/1191
https://github.com/substack/node-browserify/issues/1199
https://github.com/substack/node-browserify/issues/1213
https://github.com/substack/node-browserify/issues/1217

thedufer:
All of our build tools are carefully frozen to exact versions and we do thorough testing before pushing an upgrade to any of them.  We're perfectly willing to fix things if this behavior changes; in the meantime, it's worth depending on this UB because we're getting significant performance benefits related to caching that we wouldn't be able to get without it.

Thanks for the warning, though.

PR #1423, opened 2015-11-06
Fix paths in example of multiple bundles

glarrain:
Also replaced dir name `static` with `bundle` to make it clear those are the bundle files generated by browserify, not the source files.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -318,23 +318,23 @@ module.exports = function (s) { return s.toUpperCase() + '!' };
 ```
 
 ```
-$ browserify -r ./robot.js > static/common.js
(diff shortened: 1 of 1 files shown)


mikixing:
您的邮件我已收到，会尽快查阅~

ljharb:
this still seems relevant

glarrain:
@ljharb in case you missed it, this is a PR, not an issue, so I reckon after 10 years it's about time to close it

ljharb:
I didn't miss it, and the length of time shouldn't ever have an impact on whether a PR is still open. Since there's not an associated issue, this PR *is* the issue, and if it remains open, then someone can still land it. It wouldn't be the first decade-old PR I've landed.

Issue #1358, opened 2015-08-18
.require() with expose: true use backslashes on Windows


piranna:
When using [.require()](https://github.com/substack/node-browserify#brequirefile-opts) method on Windows with the `expose:true` flag, the IDs are being generated with backslashes, like `/apps\\Call` instead of `/apps/Call`, so later they can't be found when `require()`ing them because they are not found. I think this is somewhat related to https://github.com/substack/node-browserify/pull/503/files and https://github.com/substack/node-browserify/pull/1096/files and probably the fix would be the same, just replace the backslashes previously to set the path at the browserify output.

piranna:
Probably it would be also a good idea to sanitice this way all the paths to be sure they are only using slashes and not backslashes...

piranna:
Taking a look on the code, I suspect the problem is at [this path.resolve()](https://github.com/substack/node-browserify/blob/f7e626ca75bd925a584f9b0fe8ff9d4b8e90f188/index.js#L165). I don't have a Windows machine here to try it, though :-(

julian-maughan:
This issue is currently preventing me upgrading from a fairly old version of Browserify, so I'd be great to see it resolved. Hope the PR helps.

substack:
Should be fixed in 12.0.2.

githubVivek55:
this issue is still exist in windows for external dependencies section. Could anyone please help me how to resolve that..

PR #1424, opened 2015-11-07
Fix badly formed module paths on Windows

julian-maughan:
This addresses the issue raised in #1358.

It ensures 'required' modules have the expected path, i.e. `require("/apps/Call")` instead of `require("/apps\\Call")`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,12 +115,14 @@ Browserify.prototype.require = function (file, opts) {
     var expose = opts.expose;
     if (file === expose && /^[\.]/.test(expose)) {
         expose = '/' + path.relative(basedir, expose);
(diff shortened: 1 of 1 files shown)


anddoutoi:
Ran into a simial issue where both `browserify -r ./foo/bar.js` and `browserify -r ./foo/bar.js:./foo/bar.js` both returned some weird mix of absolute unix path and escaped windows path: `/foo\\bar.js`.

Leaving out `.js` in the target part works though, e.g. `browserify -r ./foo/bar.js:./foo/bar` returns `./foo/bar`

TheSharpieOne:
lgtm

substack:
Thanks for the patch, merged in 12.0.2.

PR #1437, opened 2015-11-19
Added failing test: browserify still transforms symlinked modules

kintel:
This is a failing test showing that #1386 wasn't properly fixed.
The problem is that while symlinked modules containing a single file works, any subsequent relative requires made from that module are still being transformed, breaking any non-trivial example.


substack:
All fixed with the patches to module-deps. Great work!

Issue #1436, opened 2015-11-18
How to provide generated content as module ?


Cactusbone:
I need to be able to provide generated content as dependencies to bundles

during build (using api) I know what the generated content will be, but there's no file containing it. it's only in memory. so i want to provide it to the bundle requiring it.

I can indeed do 
```
bundle.exclude('a');
bundle.require(stream, {expose:'a',file:'a.js'});
```

but this means my bundle ends up with 'a' even when it's not required.
(I'm using factor-bundle so having duplicated content is not welcome)
any tips ?

Cactusbone:
with custom resolve function : 

```
            var generatedModules = {
                'a': textContent,
            };
            var fileCache = {};
            _.each(generatedModules, function (val, key) {
                fileCache[key] = val;
            });

browserify({
                fileCache: fileCache,
                resolve: function (id, parent, cb) {
                    var filename = null;
                    if (generatedModules[id]) {
                        filename = id;
                    }
                    _.defer(_.partial(cb, null, filename));
                },
})
```

jmm:
@Cactusbone Perhaps what's discussed here will give you at least a temporary solution: https://github.com/substack/node-browserify/issues/1440. As noted there though, the plugin is implemented using the undocumented ability to set module-deps `.resolver`, so fundamentally it's no better than doing that (with the usual cautions about using undocumented behavior), it just provides some conveniences on top of it.

Cactusbone:
@jmm what i wrote here is working for me, so i'm using it. but since it's using a PR, i'm leaving the bug opened. 
And i've stumbled on #1440 before, but i had the file not found issue with it :/

jmm:
@Cactusbone Even if you combine it with the `fileCache` part from here?

Cactusbone:
Yeah, I'll try it again and report back :)

Cactusbone:
@jmm humm looks like you're right ! I must have found fileCache option long after trying .resolver

I like this solution better since I don't have to depend on a custom fork. even if i'm using undocumented code.

final code : 
```
var generatedModules = {
    'a': textContent,
};
var fileCache = {};
_.each(generatedModules, function (val, key) {
    fileCache[key] = val;
});

var bundle = browserify({
    fileCache: fileCache,
}
bundle.plugin(generateRequires, {generated: generatedModules});

function generateRequires(b, opts) {
    var generatedBundles = opts.generated || {};
    var deps = b.pipeline.get('deps');
    deps = deps.get(0);
[...]

PR #1441, opened 2015-11-20
accept custom resolve function

Cactusbone:
custom resolve function which default to internal resolve when not found.

allows to easily plug in some files.

not far from #1006 though

along with module_deps fileCache, can be used to provide generated content for #1436

### DIFF
--- a/index.js
+++ b/index.js
@@ -489,6 +489,19 @@ Browserify.prototype._createDeps = function (opts) {
             else cb(err, null, pkg)
         });
     };
+
+    if (opts.resolve) {
+        var baseResolve = mopts.resolve;
+        var customResolve = opts.resolve;
+        mopts.resolve = function (id, parent, cb) {
+            customResolve(id, parent, function (err, file, pkg) {
+                if (!err && !file)
+                    baseResolve(id, parent, cb);
+                else
+                    cb(err, file, pkg);
+            });
+        };
+    }
     
     if (opts.builtins === false) {
         mopts.modules = {};


jmm:
Interesting approach @Cactusbone. I've been thinking about a good way to solve that (allow setting it, but also wrapping the original). Part of the issue for me is that I wish to be able to do it from a plugin.

Cactusbone:
from what I understood, plugins can interact only after modules_deps right now, so it's not gonna be easy :)

jmm:
@Cactusbone That's true. It could be as simple as just documenting `.resolver` as public API on module-deps in combination with allowing it to be passed in as an opt. I have done some work on prototyping an extension to the plugin API that would allow plugins to run before the pipeline is constructed. I think all you'd lose is the ability to synchronously access `b.pipeline` before `b.bundle()`.

Cactusbone:
.resolver did not do the trick for me, since it still required existing files and i wanted to give arbitrary code.

jmm:
@Cactusbone I'm not sure what you mean. Wrapping `.resolver` doesn't inherently require existing files any more than this PR does, does it?

Cactusbone:
@jmm I don't remember exactly from my tests, but browser-resolve was checking if my files existed when i used .resolver,  whereas my PR changes directly in module-deps, which check first in its fileCache.

check https://github.com/substack/node-browserify/issues/1436 for my usage :)

I can check again if you want th exact stack trace

jmm:
@Cactusbone I understand what you're going for, I'm just not sure what difference there would be between what you've PR'ed here and this:

```js
var b = browserify();
var mdeps = b.pipeline.get("deps").get(0);
var baseResolve = mdeps.resolver;
mdeps.resolver = function (id, parent, cb) {
  customResolve(id, parent, function (err, file, pkg) {
    if (!err && !file) return baseResolve(id, parent, cb);
    else return cb(err, file, pkg);
  });
};
```

In your PR `opts.resolve` just becomes `.resolver`, doesn't it?

(Please note that I'm not necessarily saying your proposal shouldn't be considered -- there's a couple of different issues involved here.)

Cactusbone:
@jmm looks like my memory was playing against me. indeed using a customResolver here do work as well as my PR :)

jmm:
@Cactusbone Ok, thanks for the feedback. I personally do think there should be public API for using a custom resolver eventually -- I would just want to see it have a sensible API (including being able to set it via plugin, for example).

PR #1456, opened 2015-12-17
file is an object and set opts

wenbing:


substack:
I think `require()` probably has too many different API forms as it is. Closing unless there is a compelling reason for this use-case.

wenbing:
There is just one case, if i require an object as `{ file: filepath, expose: exposeName }` , then browserify will not set expose properly.

PR #1460, opened 2015-12-21
Add tests to .npmignore

netei:
Tests take ~ 1 Mo in space, maybe we shouldn't ship them to npm ?


substack:
The whole package is less than 64k over the wire:

```
$ curl -s 'https://registry.npmjs.org/browserify/-/browserify-12.0.1.tgz' | wc -c
63471
```

PR #1468, opened 2015-12-30
Update buffer to 3.6.0

### LINKED ISSUE #1418: Buffer 3.5.0 broken (required by XMLHttpRequest shim)
Today I ran into feross/buffer/issues/79 (Cannot set property length of [object Object]), when browserifying code, which made use of XMLHttpRequest.

(You'll need to run `npm test` on [my ajar clone](https://github.com/Ryuno-Ki/ajar/tree/mocha) after commenting line 2, where I require `xmlhttprequest`).

As stated there, the issue could be found on browserify-site as well. Please consider downgrading buffer import in [your package.json](https://github.com/substack/node-browserify/blob/b7cf5bef990a6823d14e2d437e15112aa6b3adfa/package.json#L28) until the issue is solved.

TheSharpieOne:
Resolves #1418 by using a newer version of buffer which fixes the root issue (feross/buffer#79).

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "browser-pack": "^6.0.1",
     "browser-resolve": "^1.11.0",
     "browserify-zlib": "~0.1.2",
-    "buffer": "^3.4.3",
+    "buffer": "^3.6.0",
     "concat-stream": "~1.5.1",
     "console-browserify": "^1.1.0",
     "constants-browserify": "~1.0.0",


substack:
This shouldn't be necessary since browserify floats the version of buffer with `^`.

sholladay:
To be clear: the actual solution for the problem this is trying to solve is to re-install your Node dependencies.

````sh
# Delete stale dependencies. You may need `sudo rm -rf` instead, in some situations.
rm -r node_modules;
# Delete the npm cache, because paranoia never hurts (probably not necessary).
npm cache clean;
# Do a fresh install.
npm install;
````

PR #1476, opened 2016-01-09
buffer@4

feross:
Node.js removed the `buffer.toArrayBuffer()` API sometime during node 0.11. Users who rely on it will have code that doesn't work in node.js, which is unfortunate.

`feross/buffer` v4.0.0 removed `buffer.toArrayBuffer()` to have parity with node.js.

Going forward, users can just use `buffer.buffer` to get an `ArrayBuffer` since `Buffer` is just a subclass of `Uint8Array` in both node.js and the browser.

This is a breaking change, since we're removing a method from the buffer.

I'm only sending this PR now, since buffer v4.1.0 has a 30% performance increase for lots of the buffer methods and I want users to have that :+1:

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "browser-pack": "^6.0.1",
     "browser-resolve": "^1.11.0",
     "browserify-zlib": "~0.1.2",
-    "buffer": "^3.4.3",
+    "buffer": "^4.1.0",
     "concat-stream": "~1.5.1",
     "console-browserify": "^1.1.0",
     "constants-browserify": "~1.0.0",


feross:
I would like to release this as browserify v13, if there are no objections.

substack:
Looks good to me!

feross:
Here's what's going into the changelog:

This release bumps the [`buffer`](https://www.npmjs.com/package/buffer) dependency to v4, which has one possible breaking change:

- The `buffer.toArrayBuffer()` method on `Buffer` instances has been removed.

(This is probably not a problem for 99.9% of users.)

This API was added in node v0.11.8 and
[removed before v0.12 was released](https://github.com/nodejs/node-v0.x-archive/issues/7609#issuecomment-42903457).
It was added to browserify's `buffer` implementation but was never removed when v0.12
 was released. `buffer` v4 removes it so we have full parity with node.js.

Going forward, to get an `ArrayBuffer` from a node.js-style `Buffer`, users should
just use `buffer.buffer`. This works because `Buffer` is a subclass of `Uint8Array`
in both the browser and node.js (since v3).

On the plus side, this release also includes:

- Performance improvements to `buffer` that
[increase the speed](https://github.com/feross/buffer/pull/92) of some buffer
methods by as much as 30%!

feross:
v13.0.0 is released.

Issue #1093, opened 2015-01-25
.npmignores / reducing install bloat


mattdesl:
A fresh install of browserify produces around 26.6 mb of bloat. Most of it is coming from tests, and tests of dependencies. Not a big deal for a global install, but when you are using browserify as a local dependency for WebGL/DOM testing across several dozens of modules, it starts to add up.

Some of the heavy hitters, sorted from lowest to highest in filesize:

```
384	shasum/
392	duplexer2/
416	resolve/
440	through2/
528	concat-stream/
560	buffer/
720	labeled-stream-splicer/
816	glob/
944	browser-resolve/
1176	JSONStream/
1272	browser-pack/
3144	umd/
5280	crypto-browserify/
5344	browserify-zlib/
5544	syntax-error/
7008	insert-module-globals/
12328	module-deps/
```

I could start submitting PRs for some of these, as they are easy changes.

adam-lynch:
+1

cc @4ver

sindresorhus:
:+1:

kevva:
:+1:

jeromedecoster:
:+1:

blacksonic:
start pushing the merge requests for this

substack:
Those sizes are in *bytes*:

```
$ curl -s https://registry.npmjs.org/module-deps/-/module-deps-3.7.2.tgz | wc -c
13747
```

This doesn't seem worth optimizing.

substack:
If the issue is size on disk after unpacking, 1) disk is pretty cheap 2) you can use symlinks if you've got lots of browserify copies of the same files floating around 3) you can `find -name test -or -name example -exec rm -r {} \;` after installing to save some room on disk.

substack:
I'm also not saying the slowness problem isn't real, but I think the best place to solve it will be in the npm client and registry to reduce the number of round-trips, since the size of the payload seems rather small.

maxogden:
I could be wrong, but I think this issue is more about stuff like `acorn` including it's test directory on npm.

For example, here's Disk Inventory X output on the `browserify` folder after a fresh `npm install --save` (e.g. no devDeps)

![screen shot 2015-02-22 at 1 49 29 pm](https://cloud.githubusercontent.com/assets/39759/6320687/b59276b6-ba99-11e4-82a5-1c70b7843957.png)

Here are the top offending modules in terms of on-disk usage:

[...]

PR #1486, opened 2016-01-16
package.json: use `files`

jscissr:
This makes the install faster, by not installing `test` and `example` (130KB, 428 files).
`index.js`, `LICENSE`, `changelog.markdown`, `package.json` & `readme.markdown` are automatically included, see https://docs.npmjs.com/files/package.json#files and https://github.com/feross/buffer/pull/98.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,5 +88,9 @@
   "scripts": {
     "test": "tap test/*.js"
   },
+  "files": [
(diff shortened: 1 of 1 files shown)


feross:
Duplicate of https://github.com/substack/node-browserify/issues/1093

PR #1524, opened 2016-03-16
correct typo in usage.txt

greenify:
Just stumbled across this - seems to have gone unnoticed for quite a while ;-)

### DIFF
--- a/bin/usage.txt
+++ b/bin/usage.txt
@@ -20,7 +20,7 @@ Standard Options:
  
     --command, -c  Use a transform command on top-level files.
    
-  --standalone -s  Generate a UMD bundle for the supplied export name.
+  --standalone -s  Generate an UMD bundle for the supplied export name.
                    This bundle works with other module systems and sets the name
                    given as a window global if no module system is found.
   
--- a/readme.markdown
+++ b/readme.markdown
@@ -91,7 +91,7 @@ Standard Options:
 
     --command, -c  Use a transform command on top-level files.
 
-  --standalone -s  Generate a UMD bundle for the supplied export name.
+  --standalone -s  Generate an UMD bundle for the supplied export name.
                    This bundle works with other module systems and sets the name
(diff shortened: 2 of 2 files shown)


jmm:
Hello @greenify, thanks for your interest in improving Browserify. I'm sorry :( but I don't consider that a typo and favor keeping it as-is. I've never understood it when people want to write things that way, because I would never say it that way and this is a case where I think it makes sense to write it the way it would be pronounced. But again, thank you for contributing :) I'm going to leave this open for now in case others want to comment.

greenify:
>  I would never say it that way and this is a case where I think it makes sense to write it the way it would be pronounced

I am afraid I can't follow your reasoning: we are talking about the English language - not a software standard.
AFAIK the rule is "[the form an is used before words starting with a vowel sound](https://owl.english.purdue.edu/owl/resource/540/01/)", however I am not a native speaker. Isn't @substack one?

> thank you for contributing 

Yeah I love open source anyways and it's not so important too me whether you fix this trivial issue - don't worry ;-)

greenify:
> I am afraid I can't follow your reasoning: we are talking about the English language - not a software standard.

Sorry about being so grumpy - I just looked it up in more dictionaries and apparently "u" in this case is considered to be a consonant sound - one also says "a UNICEF" member. You have a weird language ;-)

terinjokes:
@greenify Native speaker here. One of the rules "taught" to us is that in the English language there are exceptions to every rule (and I'm sure even this rule has an exception).

From Purdue OWL's article ["Articles: A versus An"](https://owl.english.purdue.edu/owl/resource/591/01/):

> ## Exceptions
> When "u" makes the same sound as the "y" in "you," or "o" makes the same sound as "w" in "won," then a is used.

Since "u" in "UMD" is pronounced, well, like "you", the correct article is "a".

But again, thanks for your contribution to the Browserify docs. :smile:

Issue #1543, opened 2016-04-13
Caching the results of path.relative reduces rebuild times


ashaffer:
I did a bit of profiling with [devtool](https://github.com/Jam3/devtool) and noticed that a bunch of time is spent in `path.relative` during rebuilds, for the most part repeating path resolutions that have already been done.

Memoizing it's results decreases my rebuild times by about 10%. I made a little module for shimming path.relative to make this happen, [cached-path-relative](https://github.com/ashaffer/cached-path-relative). But maybe it'd be best to have browserify and its dependencies use this or something like it explicitly?

terinjokes:
I love when 🔍  happens 😄 .

Other than removing the repeated relative lookups entirely, I don't see why we wouldn't want to cache these. I can see a problem is a symlink changes, but Browserify doesn't have a great record with those already. 😞 

Where would this make the most impact in your builds? Here in browserify, or in one of the dependencies?

ashaffer:
The big ones seem to be here in browserify and in module-deps. And ya, devtool is pretty awesome for this sort of profiling.

terinjokes:
PR? 😀

ashaffer:
Sure, i'll try to put something together tonight.

substack:
Great work everyone. Released in 13.1.1.

PR #1544, opened 2016-04-14
use memoized path.relative

ashaffer:
Per #1543 @terinjokes


substack:
Thanks for the patch! Merged in 13.1.1 with whitespace changes removed.

PR #1551, opened 2016-05-03
if _options.basedir exists, use it in external, exclude and ignore

tufandevrim:
Not sure if it was intentional or not, main option basedir (this._options.basedir) is being ignored in external, exclude and ignore functions. It is being taken into account in require function by the way https://github.com/substack/node-browserify/blob/master/index.js#L114


jmm:
Hello @tufandevrim, thanks for contributing. At a glance this does seem to make sense, but it'd help to have failing tests that show the current implementation doing the wrong thing.

(Side note: totally not your fault, but I noticed that these all have an undocumented `opts` param and read at least `opts.basedir` from it...ugh.)

tufandevrim:
Hi @jmm 
Please take a look at the test I just added and let me know if it makes sense or not.
If you run this test without the changes proposed in this PR, it will fail unless you add basedir option explicitly to this line https://github.com/substack/node-browserify/pull/1551/files#diff-f101fe74380a155175b7b9240f1b1fcaR15

jmm:
Thanks @tufandevrim! I or someone will take a look when there's a chance.

pushred:
This threw me off today. Because of the undocumented opts param I assumed that `this._options` was the one to use.. I'd submit a PR for a doc update but seems this is the better approach if `require` already does this.

PR #1552, opened 2016-05-04
Use native Array.isArray

kt3k:


jmm:
Hello @kt3k, thanks for contributing. LGTM -- I'll merge, but before I do a couple things:

* How about instead of repeating `Array.isArray` we do this:

  ```diff
  - var isarray = require('isarray');
  + var isArray = Array.isArray;
  ```

  Then of course change all `isarray => isArray`.

* I see that you retained `isarray` in `package.json` and bumped the version. Did you try to determine whether there's anything else still using that dep? Maybe we can just get rid of it now.

Notes: according to my research [v8 added `Array.isArray` in v1.3.16]( https://github.com/v8/v8/blob/fd4f63b652305646c10c779a010b19edba9ca536/ChangeLog#L14460), which would make it available from [Node v0.1.15](https://github.com/nodejs/node/blob/a4f94b427170bdef44a4c4ba45ff36664fab6597/doc/CHANGELOG.ARCHIVE.md#20091028-version-0115).

kt3k:
> var isArray = Array.isArray;

OK. That looks better. I'll change it.

> I see that you retained isarray in package.json and bumped the version. Did you try to determine whether there's anything else still using that dep? Maybe we can just get rid of it now.

I keep `isarray` in `devDependencies` because it's used in test/array.js as an example of bundleing of npm modules. It seems purely used as an example of npm module, and I think it's ok to keep it (because it only affects tests).

jmm:
@kt3k Thanks!

> I keep `isarray` in `devDependencies` because it's used in test/array.js as an example of bundleing of npm modules. It seems purely used as an example of npm module, and I think it's ok to keep it (because it only affects tests).

Oh ok, sorry, I hadn't noticed that you moved it to devDeps. Nonetheless, if that's the only thing it's used for I'm going to see if there's something else that we're depending on for some other reason that can be substituted so we can just eliminate `isarray`.

kt3k:
I replaced test cases which use 'isarray' by ones using 'xtend' instead, and removed 'isarray' from devDeps. Looks cleaner!

jmm:
@kt3k Thanks! I had actually done that part myself locally yesterday but hadn't had the chance to push yet. I took a look at your latest commit and I certainly appreciate you doing that, but for that part I used what I committed yesterday only because it's a smaller diff. I replaced `isarray` with `defined` which ended up with the same assertion methods and number of assertions (and even `expected` value!).

I squashed your first 2 commits together and merged via #1555. Thanks again!

kt3k:
My pleasure! Thanks for merging 😄 
Appreciate your quick response!

jmm:
Published in 13.0.1. Not to discourage you from contributing, but just FYI -- responses usually aren't so quick.

PR #1581, opened 2016-06-23
Add option to turn off deduping

ganemone:


sindresorhus:
You should probably elaborate on **why**.

ganemone:
What initially prompted me to do this was an issue with the livereactload library and things being incorrectly deduped. In general though, I think it is useful for debugging. Deduping is not an easy thing to do (just look at npm3 for example). When weird things are happening with builds, it would be useful to be able to turn off deduping to help diagnose/isolate the issue.

ganemone:
Sorry! I think I accidentally closed the PR when I commented on it last time. Still interested in getting this merged if it looks good.

substack:
This patch looks good to me, it merely exposes features that are already available in sub-modules.

PR #1584, opened 2016-06-29
Replace \\ with / in relative external paths

ale316:
Currently files required from other bundles imported like follows in Windows:
```
  bundler.external(['./assets/Master.js']);
```
will produce require calls as follows in the compiled file:
```
  var Master = require('/assets\\Master.js');
```
which will naturally fail. This fix replaces the double slashes with a single slash, like it's done in `Browserify.prototype.require`

### DIFF
--- a/index.js
+++ b/index.js
@@ -253,8 +253,10 @@ Browserify.prototype.external = function (file, opts) {
     
     if (!opts) opts = {};
     var basedir = defined(opts.basedir, process.cwd());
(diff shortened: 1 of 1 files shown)


ale316:
Tests pass on my machine:
screen shot 2016-06-29 at 15 52 27

Oscar-ren:
excuse me , could you tell me how do you solve this problem？thanks，I don't see the master merge this commit.

PR #1611, opened 2016-08-10
deps: update assert

jscissr:
This includes new methods such as `deepStrictEqual`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   ],
   "dependencies": {
     "JSONStream": "^1.0.3",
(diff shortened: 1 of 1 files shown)


jscissr:
On my Windows machine, I have 59 tests failing because of Windows paths, but *not* the two failing on travis. Can someone not on Windows investigate please? I have no clue why the test fails, assert does not appear anywhere in them.

jscissr:
It looks like all I had to do was restart the build. Strange.

PR #1620, opened 2016-08-31
update tests to use Buffer.from

MylesBorins:
Using Buffer without new will soon stop working. It will throw
A deprecation warning in Node v7. This commit modifies the tests
in `test/bare.js` and `test/bare_shebang.js` to use the
`Buffer.from` interface instead. It is a drop in replacement for
Buffer that is supported by Node v4 / v6 and Browserify's Buffer
implementation

For the browserify test suite to pass on Node v7 we will need
to land this commit as well as the two below PR's
- https://github.com/substack/browser-pack/pull/75
- https://github.com/maxogden/concat-stream/pull/48


TheAlphaNerd:
It is worth mentioning that this breaks the test suite on node v0.10 and v0.12. Alternatively we can use the `new Buffer` interface to do this instead, the transform function for lower casing will have to be modified though

substack:
This upstream behavior was reverted in node 7, so closing for now.

Issue #1616, opened 2016-08-19
`Buffer()` without a `new` keyword is going to be hard-deprecated soon


ChALkeR:
Refs: https://github.com/nodejs/node/pull/8169, https://github.com/nodejs/node/pull/7152.

Two options here:
- Use the new Buffer API — `Buffer.alloc()`/`Buffer.from()`/`Buffer.allocUnsafe()` (requires [a shim](https://github.com/feross/safe-buffer) for v0.10/v0.12 and older v4.x versions prior to v4.5.0).
- Use `new Buffer()` for the time being — just add the `new` keyword everywhere. You should manually check that everything is safe in this case, see the links below for more explanation. That _might_ be hard-deprecated at some later point.

More background:
- https://github.com/ChALkeR/notes/blob/master/Lets-fix-Buffer-API.md
- https://github.com/ChALkeR/notes/blob/master/Buffer-knows-everything.md
- https://github.com/nodejs/node/pull/7152#issuecomment-240880317

Quick grep (you should better re-check):

```
browserify-13.1.0.tgz/test/bare.js:34:    ps.stdin.end('console.log(Buffer("ABC"))');
browserify-13.1.0.tgz/test/bare_shebang.js:32:    ps.stdin.end('#!/usr/bin/env node\nconsole.log(Buffer("WOO"))');
browserify-13.1.0.tgz/test/global/buffer.js:1:t.equal(Buffer('xyz').toString('base64'), 'eHl6');
browserify-13.1.0.tgz/test/global/buffer.js:2:t.equal(Buffer('eHl6', 'base64').toString(), 'xyz');
browserify-13.1.0.tgz/test/leak.js:46:    stream.push('t.equal(Buffer("eHl6", "base64").toString(), "xyz")');
```

The grep above only includes the lines that call `Buffer()` without the `new` keyword, if you choose to move to the new API — you should probably also replace `new Buffer(…)` calls.

terinjokes:
😑 . Time to make sure the entire browserify dependency tree is updated, I suppose.

theskillwithin:
i just installed browserify,  nodejs 7.0

its a warning, but is it even running?

```
> chiceventsweddings@0.5.0 babel /Users/Austin/Projects/chicevents
> browserify src/assets/js/functions.js -t babelify --out-file dist/assets/js/app.js --source-maps

[...]

PR #1647, opened 2016-11-05
Fix usage of deprecated Buffer API

seishun:
```
(node:23274) DeprecationWarning: Using Buffer without `new` will soon stop working. Use `new Buffer()`, or preferably `Buffer.from()`, `Buffer.allocUnsafe()` or `Buffer.alloc()` instead.
```

See #1616. Only tests were affected.

Once Node.js 0.12 reaches EOL (January 1 2017), or whenever you feel comfortable with dropping 0.12 support, you can safely remove the "safe-buffer" dependency and the `require`s.


mourner:
@substack any chance of merging this harmless PR? Those Node 7 warnings are pretty annoying.

substack:
This patch updates tests. Why?

substack:
I don't like this.

seishun:
You don't like this because it updates tests?

>This patch ONLY updates tests, which I also don't understand.

"Only tests were affected." https://github.com/substack/node-browserify/pull/1647#issue-187509519

PR #1670, opened 2017-01-02
Allow configuring transformKey option for module-deps

laurynasl-wix:
module-deps supports specifiying the transformKey to use for looking up the transforms defined in package.json of dependencies. In some cases it is desirable to suppress this behaviour. This PR allows you to specifiy this via browserify options.


substack:
Thanks for the patch. Merged into 13.2.0.

PR #1689, opened 2017-02-15
Remove mention to browserifysearch

oncletom:
The website seems down or not active anymore.


goto-bus-stop:
yeah, that's confusing. thanks!

Issue #1629, opened 2016-09-28
Referencing an exposed module by its path fails with fullPaths option


niallwingham:
Minimal reproduction: https://github.com/niallwingham/browserify-full-paths-bug

There is a bug caused by the `fullPaths` option, when one of the bundled modules requires an exposed module using a path.  The exposed module is available under its alias, but not under its file path.  The bug is avoided when `fullPaths` is not used because both the file path and the alias are mapped to an index by deps-sort.

I'm happy to contribute a pull request to resolve this.  It could be fixed by doing a similar mapping in deps-sort when the `index` option is false, if we want to address it there.  But I'm not sure if it should be the job of deps-sort to correct dependency mappings; maybe this belongs in module-deps or elsewhere.  Please advise!

Issue #1534, opened 2016-03-23
unable to emit full absolute file paths for dependencies


mstorus:
I am unable to make browserify output full absolute file paths for modules. Is this possible?

For example, I expected browserify to produce full file paths as dependency keys when executing:
```
browserify -x /Users/me/code/a.js bundle.js
```

given the following files:
```
$ cat a.js 
module.exports = function () {return 'aaaa';}
$ cat bundle.js
var a = require('./a.js');
console.log(a());
```

But instead the output dependencies of `bundle.js` from browserify look like this:
```
{"./a.js":"/a.js"}]},{},[1]
```

Is it possible to make browserify output the full file path like this?:
```
{"./a.js":"/Users/me/code/a.js"}]},{},[1]
```

Looking through https://github.com/substack/node-browserify/blob/8f7adc96e3ed59c299d078405a463860f5409ba6/index.js#L695-L702,
this code seems strange:
```
            if (self._external.indexOf(rfile) >= 0) {
                row.deps[key] = rfile;
            }
            if (self._external.indexOf(afile) >= 0) {
                row.deps[key] = rfile;
            }
```

whether we find the relative file (`rfile`) or absolute file (`afile`) in `_external`, we always set `row.deps[key]` to `rfile`. Why don't we set `row.deps[key]` to `afile` ever?

Perhaps it's something to do with how we append both the relative file path and the file path itself to `_externals`:
https://github.com/substack/node-browserify/blob/8f7adc96e3ed59c299d078405a463860f5409ba6/index.js#L255-L257
```
    this._external.push(file);
    this._external.push('/' + path.relative(basedir, file));
```

But I don't really understand why this is done.

mderijcke:
Are you using the `--full-paths` option (`opts.fullPaths = true` using task runners / code)? It produces this for me on Windows:

```
},{"./a.js":"C:\\Users\\Mitchell\\Desktop\\test\\a.js"}]},{},["C:\\Users\\Mitchell\\Desktop\\huh\\test.js"]);
```

Which appears to be the desired behavior, right?

mderijcke:
Reading your question thoroughly, I think this is desired behavior. You'll always going to want to reference files relatively from your bundle, which in most cases should stay within the project, and in worst case scenario get you a ton of `../../../../`.

The if statement is a little bit awkward indeed, as it behaves the same in both statements. But if you want full paths, I still suggest just turning on `--full-paths` but generally you'll not going to want to expose them.

PR #1690, opened 2017-02-16
Fix interaction between fullPaths and external

MellowMelon:
Consider the case where you are splitting bundles with your own code, not in node_modules. You have bundle 1, which has the `require` calls, and you also have bundle 2, which makes the same `external` calls. Assume the default expose names and relative require arguments with no aliasing magic.

When `fullPaths` is false in both, this works fine if you are careful with setting `basedir`. When `fullPaths` is true in both, bundle 1 will expose absolute paths but bundle 2's external IDs are the ones relative to the basedir. As #1534 notes, currently you can't get bundle 2 to look for the absolute paths, so if you want to use an ecosystem tool requiring `fullPaths`, you're stuck.

(The potential workaround of specifying `expose` yourself may run into #1629, though I didn't check carefully. For that issue, I think the suggested fix works, i.e. patch deps-sort in the obvious way, but that's separate from this PR.)

This PR makes it so that if you compile bundle 2 with `fullPaths` true, the external IDs are absolute paths. So `fullPaths` is now usable. One minor disadvantage: consistent `fullPaths` settings across bundles is now a requirement. If someone has it where bundle 1 is compiled with `fullPaths` false and bundle 2 is compiled with `fullPaths` true, their setup works now and my PR will break it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -697,11 +697,12 @@ Browserify.prototype._label = function (opts) {
 
             var afile = path.resolve(path.dirname(row.file), key);
             var rfile = '/' + relativePath(basedir, afile);
+            var exposefile = opts.fullPaths ? afile : rfile;
             if (self._external.indexOf(rfile) >= 0) {
-                row.deps[key] = rfile;
+                row.deps[key] = exposefile;
             }
             if (self._external.indexOf(afile) >= 0) {
-                row.deps[key] = rfile;
+                row.deps[key] = exposefile;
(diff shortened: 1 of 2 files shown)


MellowMelon:
Nevermind. Some further testing suggests things aren't generalizing the way I thought they did when I submitted this. I'll reopen if I think this fix still makes sense after I work out the details.

PR #1704, opened 2017-03-20
Fix external relative paths on Windows

Shingyx:
Replace backslashes with forward slashes when adding relative paths for external references.


Shingyx:
Fixes the following scenario:

- Set up bundle A which requires './lib/foo.js', exposed at 'lib/foo'.
- Set up bundle B which includes a src file './test/bar.js' and marks 'lib/foo' as an external dependency.
- If './test/bar.js' has the statement `require('../lib/foo')`, it'll throw the error something along the lines of Cannot find module '/lib\foo'.

Shingyx:
Build is failing for the same reason that the master branch is failing https://travis-ci.org/substack/node-browserify/builds/207173302

Shingyx:
[grunt-test.zip](https://github.com/substack/node-browserify/files/857317/grunt-test.zip)
Attached is a small test project which reproduces this problem.

Shingyx:
...I may have just had a misunderstanding of how Browserify works. I will have to verify tomorrow, but it seems like my change just gave me the illusion of fixing the problem, when in reality it was just including './lib/foo.js' in the second bundle, so my change probably breaks the functionality of the external flag and I was misusing Browserify the whole time. Will close if this is the case.

Shingyx:
Can confirm. All my change seemed to do was invalidate the `external` option, making the second bundle just include the "external" files. However, I think my scenario suggests that there _could_ be an issue somewhere in Browserify's external option, given the error `Cannot find module '/lib\foo'` was thrown, but my change does nothing to address this issue (if there is one at all).

dwelle:
I can confirm the backslashes on windows are causing problems. Why did you conclude your proposed change is incorrect?

Related https://github.com/substack/node-browserify/pull/1584, which was not resolved either

Shingyx:
Firstly, ignore my previous grunt-test.zip. It was more complicated than necessary.

[grunt-test.zip](https://github.com/substack/node-browserify/files/874630/grunt-test.zip)
This is a much simpler project which demonstrates the problem. For context, I want to bundle a bunch of production code inside `lib-bundle.js` and the test code inside `test-bundle.js`. For those tests, I want the dependencies to come from `lib-bundle.js`, AKA _exactly_ what will be used in production. I use `index.html` to run those tests (usually would be using Mocha or something).

[dist-before.zip](https://github.com/substack/node-browserify/files/874636/dist-before.zip)
This zip contains those two bundles before my changes. This is everything I wanted, _except_ the last line of `test-bundle.js` has `{"../lib/library":"/lib\\library.js"}]}`, saying 'to import "../lib/library", look up "/lib\\library.js"'. As a result, opening `index.html` will give "Uncaught Error: Cannot find module '/lib\library.js'". Simply replacing the `\\` on this line with `/` gives the expected behaviour since "/lib/library.js" is declared in `lib-bundle.js`, which is what I was aiming to do.

[dist-after.zip](https://github.com/substack/node-browserify/files/874646/dist-after.zip)
This zip contains the bundles after my changes. `lib-bundle.js` is unchanged, but `test-bundle.js` just decided to import the files I was trying to mark as `external` in itself, suggesting that all my changes did was invalidate the `external` option.

So while I think there _is_ a problem in this use case, my changes didn't fix it.

dwelle:
There's no doubt there _is_ a problem (as the numerous issues attest). What I'm afraid of that there's not any incentive to fix it (there were multiple PRs somewhat addressing windows backslash issues, [some](https://github.com/substack/node-browserify/pull/740) still open).

/cc @substack should we attempt to issue a PR to fix this at all?

Shingyx:
Fixed it properly this time by replacing all \ with / in relative paths. Now it works as expected.

Fun fact: my change decreased the number of failing tests on my machine from 61 to 53.

Shingyx:
Is there anything else I can do to help get this merged in?

kauvj:
I am also facing same issue. Will this get merged?

Shingyx:
@kauvj I haven't heard anything, but I merged in the current master branch to get this to build.

index.js
```diff
@@ -118,14 +119,12 @@ Browserify.prototype.require = function (file, opts) {
     var expose = opts.expose;
     if (file === expose && /^[\.]/.test(expose)) {
         expose = '/' + relativePath(basedir, expose);
-        expose = expose.replace(/\\/g, '/');
```
mafintosh: Couldn't you use your helper function from below here instead?

index.js
```diff
@@ -118,14 +119,12 @@ Browserify.prototype.require = function (file, opts) {
...
     }
     if (expose === undefined && this._options.exposeAll) {
         expose = true;
     }
     if (expose === true) {
         expose = '/' + relativePath(basedir, file);
-        expose = expose.replace(/\\/g, '/');
```
mafintosh: Same as above

index.js
Shingyx: I'm not sure what you mean. My changes basically extended the existing `relativePath` method to replace backslashes with forward slashes, which means that adding the `.replace(/\\/g, '/')` after a `relativePath` call is no longer necessary.

I don't quite remember, but I think the problem was that `.replace(/\\/g, '/')` was missing after some of the other `relativePath` calls and so the path behaviour was inconsistent on Windows.

goto-bus-stop:
Will merge this in #1819. Thanks!

Shingyx:
Awesome. Thanks!

dwelle:
Still not fixed entirely as it prepends an extra `/` at the start of the path, see https://github.com/browserify/browserify/issues/1511 for more info.

Issue #1706, opened 2017-03-29
browserify field flags to switch behavior based on dev/prod/...


serapath:
Browserify has the awesome feature, that every dependency can specify it's own transforms in a "browserify" field in it's `package.json` :+1: 

**My issue:**
* While I'm developing I run watchify and don't need minification or extensive testing or transpilation of modern es6 code and beyond and I want my pipelin to be fast.
* When I'm building for production, I want minification and transpilation es6toes5 for legacy browser support and a separate source maps file and so on...

**Idea:**
Is it or would it be possible to switch the behavior of a module's "browserify pipeline" as defined in it's "browserify field" based on environment variables or alternatively based on spec?

**I would be willing to try to come up with an implementation and a pull request - but i first want to make sure if this would be acceptable and if so - what would be the best way** :-)

MellowMelon:
To ignore the transform/browserify fields of other package.json files, pass `transformKey: false` when constructing the browserify object.

This was very recently added and apparently not documented. Check the [last commit](https://github.com/substack/node-browserify/commit/bf061d8d8cd7769c3a74ef4569d7adda60ef3f4f) that changed this repository's index.js. That commit also has a test you can look at.

serapath:
I don't want to ignore the transform of browserify fields of other package.json files. It's awesome and i like it.

I just would like to customize the transforms listed in each projects package.json based on environment variables, e.g. production, development, staging, ...

For example, in `"production"` i want to do minification or polyfilling or a lot of more things that i might need in order to support old browsers and improve performance, etc...

In  `"development"` i dont care to minify or to polyfill, because i'm developing on a modern browser. So i'd like to skip a lot of transforms to speed up the build process, because i want to live-reload on every file save and that should be fast.

babel for example seems to support something like `env options` https://babeljs.io/docs/usage/babelrc/

Otherwise, I am just wondering if such a feature would fit with browserify and if not if maybe things can be done differently or in other places, so that there is no need for those flags.

serapath:
Here you go - changedt he proposal from above slightly to use:
* `browserify index.js --transform-key=production > bundle.js`

This is probably better than camelcase...

serapath:
ok got merged. perfect :-)

PR #1709, opened 2017-04-04
add cli support for --transform-key to support mode's like production…

serapath:
…/staging/etc..

Details see: https://github.com/substack/node-browserify/issues/1706#issuecomment-291543517

### DIFF
--- a/bin/advanced.txt
+++ b/bin/advanced.txt
@@ -56,6 +56,15 @@ Advanced Options:
     Turn off package.json browser field resolution. This is also handy if you
     need to run a bundle in node.
 
+  --transform-key
+
+    Instead of the default package.json#browserify#transform field to list
+    all transforms to apply when running browserify, a custom field, like, e.g.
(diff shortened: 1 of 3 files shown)


serapath:
Can this be merged anyway even though tests are failing?
This seems to be not related to my pull request at all.

Only `bin/advanced.txt`, `readme.markdown` and `bin/args.js` were touched.
In `bin/args.js` it's just one more command line option - so i don't see how this can possibly responsible for the breaking test.

serapath:
Maybe it would be even better to name it `--mode` or short `-m` instead, because `--transform-key` is kind of long.

serapath:
@substack another 2 pull request to fix a problem with dependencies that do not specify the alternative `--transform-key`. They affect `browserify` and `module-deps`.
* https://github.com/substack/node-browserify/pull/1721
* https://github.com/substack/module-deps/pull/131

PR #1710, opened 2017-04-04
update https-browserify to ^1.0.0

feross:
See: https://github.com/substack/https-browserify/commit/7295a7cf5c0928f8fd7e24ed8aae17801726e88b

Match Node.js behavior exactly

- Remove unused 'scheme' property
- Throw exception if non-https URL is passed in
- Support string URL arguments
- First argument is always required, so remove code that sets default
object if params doesn't exist

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "glob": "^7.1.0",
     "has": "^1.0.0",
     "htmlescape": "^1.1.0",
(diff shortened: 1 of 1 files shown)


feross:
Tests are failing due to https://github.com/nodejs/nodejs.org/issues/1191, so merging.

PR #1714, opened 2017-04-07
Updated string_decoder to v1.0.0

mcollina:
As titled.


feross:
Tests pass with no issues. Are there are user-visible breaking changes? @mcollina Should this be semver minor or major, in your opinion?

mcollina:
There are no visible breaking change. We bumped it in readable-stream with no breakage whatsoever. A minor release is ok.

feross:
14.4.0

PR #1722, opened 2017-05-11
Document ignoreMissing option for browserify() (closes #1595)

### LINKED ISSUE #1595: Programmable API alternative to --ingore-missing?
When using the Browserify CLI to bundle files, you can use the `--ignore-missing` flag to ignore all missing dependencies in the file. 

What is the programmable API alternative to that? The `opts` object doesn't seem to have an alternative either, and I couldn't find anything in the documentation. 

Thanks.

blahah:
The option is used by the CLI but was previously not documented in the README.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -460,6 +460,9 @@ When `opts.detectGlobals` is true, scan all files for `process`, `global`,
 `__filename`, and `__dirname`, defining as necessary. With this option npm
 modules are more likely to work but bundling takes longer. Default true.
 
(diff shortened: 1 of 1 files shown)


MellowMelon:
I'm guessing the original intent was that it would be in the https://github.com/substack/module-deps docs, which is where the option is actually used. There is a line in the documentation here that says "All other options are forwarded along to module-deps and browser-pack directly."

That's awfully obtuse though, so I'll merge this. Thanks.

Issue #1723, opened 2017-05-17
Using __dirname and `--node` causes absolute paths in the final bundle


jviotti:
Consider `foo.js`:

```js
console.log(__dirname);
```

Running `browserify foo.js --node` results in:

```js
(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
(function (__dirname){
console.log(__dirname);

}).call(this,"/Users/jviotti/Projects/playground/node-browserify")
},{}]},{},[1]);
```

Notice how an absolute path (`/Users/jviotti/Projects/playground/node-browserify`) from the computer the bundle was generated from, was hard-coded in the bundle. This means that taking the bundle to another machine will probably break the code.

As a solution, if `--node` is passed, we can use the final bundle's `__dirname` to dynamically resolve each file's `__dirname`.

jviotti:
Looks like the option that causes the issue is really `--bare`, which is set when passing `--node`

jviotti:
And the culprit seems to be `insert-module-globals` (https://github.com/substack/insert-module-globals).

```js
// See https://github.com/substack/insert-module-globals/blob/master/index.js#L47
...
    __filename: function (file, basedir) {
        var filename = '/' + path.relative(basedir, file);
        return JSON.stringify(filename);
    },
    __dirname: function (file, basedir) {
        var dir = path.dirname('/' + path.relative(basedir, file));
        return JSON.stringify(dir);
}
...
```

PR #1725, opened 2017-05-17
Dynamically calculate __dirname and __filename when --node is passed

jviotti:
When `--no-commondir` is set (as a consequence of `--node` for example),
it will cause the final bundle to hardcode absolute values of the
machine that generated the bundle in order to resolve `__dirname` and
`__filename`. For example, consider a `foo.js` file containing:

```js
console.log(__dirname);
console.log(__filename);
```

Calling `browserify --node` on it results in:

```js
(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
(function (__filename,__dirname){
console.log(__dirname);
console.log(__filename);

}).call(this,"/Users/jviotti/Projects/playground/node-browserify/foo.js","/Users/jviotti/Projects/playground/node-browserify")
},{}]},{},[1]);
```

Notice the absolute paths at the end of the bundle. This means that
Browserify node users can't generate a bundle in one machine, and expect
it to run without issues on another machine.

As a solution, we can use the final bundle's `__dirname` to dynamically
resolve every file's `__dirname` and `__filename`. This change only
takes place when `--node` is set, and keeps Browserify backwards
compatible.

For example, the above output might contain the following line:

```
}).call(this,require("path").join(__dirname,"foo.js"),require("path").join(__dirname,"."))
```

Fixes: https://github.com/substack/node-browserify/issues/1723
See: https://github.com/resin-io/etcher/issues/1429
See: https://github.com/resin-io/etcher/pull/1409
Signed-off-by: Juan Cruz Viotti <jviotti@openmailbox.org>

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,4 @@
+var path = require('path');
 var mdeps = require('module-deps');
 var depsSort = require('deps-sort');
 var bpack = require('browser-pack');
@@ -551,6 +552,19 @@ Browserify.prototype._createDeps = function (opts) {
                 return through();
             }
         }
+
+        if (opts.commondir === false && opts.builtins === false) {
+          opts.insertGlobalVars = xtend({
+            __dirname: function(file, basedir) {
+              var dir = path.dirname(path.relative(basedir, file));
+              return 'require("path").join(__dirname,' + dir.split(path.sep).map(JSON.stringify).join(',') + ')';
+            },
+            __filename: function(file, basedir) {
+              var filename = path.relative(basedir, file);
+              return 'require("path").join(__dirname,' + filename.split(path.sep).map(JSON.stringify).join(',') + ')';
+            }
+          }, opts.insertGlobalVars);
+        }
         
         var vars = xtend({
             process: function () { return "require('_process')" },
@@ -560,11 +574,11 @@ Browserify.prototype._createDeps = function (opts) {
             vars.process = undefined;
             vars.buffer = undefined;
         }
-        
+
         return insertGlobals(file, xtend(opts, {
             debug: opts.debug,
             always: opts.insertGlobals,
-            basedir: opts.commondir === false
+            basedir: opts.commondir === false && isArray(opts.builtins)
                 ? '/'
                 : opts.basedir || process.cwd()
             ,
--- a/test/bare.js
+++ b/test/bare.js
@@ -3,6 +3,10 @@ var spawn = require('child_process').spawn;
 var path = require('path');
 var concat = require('concat-stream');
 var vm = require('vm');
+var fs = require('fs');
+var temp = require('temp');
+temp.track();
+var tmpdir = temp.mkdirSync({prefix: 'browserify-test'});
 
 test('bare', function (t) {
     t.plan(4);
@@ -41,14 +45,17 @@ test('bare', function (t) {
 test('bare inserts __filename,__dirname but not process,global,Buffer', function (t) {
     t.plan(2);
     
+    var file = path.resolve(__dirname, 'bare/main.js');
     var ps = spawn(process.execPath, [
         path.resolve(__dirname, '../bin/cmd.js'),
-        path.resolve(__dirname, 'bare/main.js'),
+        file,
         '--bare'
     ]);
     
     ps.stdout.pipe(concat(function (body) {
         vm.runInNewContext(body, {
+            require: require,
+            __dirname: process.cwd(),
             console: {
                 log: function (msg) {
                     t.same(msg, [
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -551,6 +552,19 @@ Browserify.prototype._createDeps = function (opts) {
...
         }
+
+        if (opts.commondir === false && opts.builtins === false) {
+          opts.insertGlobalVars = xtend({
+            __dirname: function(file, basedir) {
+              var dir = path.dirname(path.relative(basedir, file));
+              return 'require("path").join(__dirname,"' + dir.split(path.sep).join('","') + '")';
+            },
+            __filename: function(file) {
+              var filename = path.relative(basedir, file);
```
lurch: Possibly-stupid first question:
How come `__dirname` has the parameters `(file, basedir)`, but `__filename` only has the parameter `file`? In that case, where does the `basedir` variable in this line come from?

test/bare.js
```diff
@@ -68,3 +75,72 @@ test('bare inserts __filename,__dirname but not process,global,Buffer', function
         t.equal(code, 0);
     });
 });
+
+test('bare inserts dynamic __filename,__dirname', function (t) {
```
lurch: Possibly-stupid second question:
The test description mentions both `__filename` and `__dirname`, but AFAICT this is only testing `__dirname`?

index.js
jviotti: Oops, fixed!

test/bare.js
jviotti: These tests execute the following file: https://github.com/substack/node-browserify/pull/1725/files#diff-df4230a729c19a353d1405d5459df60d, which prints both `__dirname`, and `__filename` (and we assert the result of both)

index.js
lurch: So how come this wasn't being picked up by the tests? Does it imply that it'd be worth testing `__dirname` and `__filename` separately, rather than testing them together?

jviotti: Looks like there is another `basedir` variable defined above (https://github.com/substack/node-browserify/blob/master/index.js#L428):

```js
var basedir = defined(opts.basedir, process.cwd());
```

And that happens to have the right one as well (its basically what gets passed to the `__filename` function).

I'll see if I can somehow break it and test it.

jviotti: I give up. The variable declared above is always valid, and I can't find a way to break it. I guess a linter should be the correct way to detect these kinds of subtle issues.

lurch: Yeah, I guess only a linter is able to automatically spot 'conflicting' variable names and suggest that one of them gets renamed, but that's obviously well outside the scope of this PR ;-)

inDream:
@substack @feross When will this fix be merged?

AKK9:
Is it possible to have this fixed?

index.js
```diff
@@ -551,6 +552,19 @@ Browserify.prototype._createDeps = function (opts) {
                 return through();
             }
         }
+
+        if (opts.commondir === false && opts.builtins === false) {
+          opts.insertGlobalVars = xtend({
+            __dirname: function(file, basedir) {
+              var dir = path.dirname(path.relative(basedir, file));
+              return 'require("path").join(__dirname,"' + dir.split(path.sep).join('","') + '")';
```
goto-bus-stop: paths _can_ contain " on some filesystems, could you use JSON.stringify here instead of manually wrapping the path in quotes?

index.js
```diff
@@ -551,6 +552,19 @@ Browserify.prototype._createDeps = function (opts) {
                 return through();
             }
         }
+
+        if (opts.commondir === false && opts.builtins === false) {
+          opts.insertGlobalVars = xtend({
```
goto-bus-stop: could this be done in cmd/args.js, in the same place as the other `node:` option stuff?

index.js
jviotti: Sounds good, done!

index.js
jviotti: Not sure I understand this comment. Do you mean setting `commondir: false` and `builtins: false` if `node: true`, and then just check for `node: true` here?

goto-bus-stop: oh sorry nevermind, i misunderstood what this check is doing--i thought you were using this to essentially check if `--node` was passed, but `opts.commondir` actually is the one causing the issue.

jviotti:
Awesome, thanks a lot and sorry for the delay addressing the comments!

jviotti:
@goto-bus-stop 

> do you think we need a major version bump for this?

Since all the previous tests are passing, this is a backwards compatible change, which I guess should be minor version bump.

PR #1734, opened 2017-06-08
Bump `os-browserify`

kevva:
This adds support for `os.homedir()`.


goto-bus-stop:
thanks! this was added in #1767 and released in 14.5.0, sorry for the wait!

PR #1742, opened 2017-06-29
Add support for the Node v6 preserve-symlinks option

devongovett:
When specified, symlinks are preserved when resolving modules. Specifically, the `realpath` call is skipped. This allows symlinked peer dependencies to work correctly. Matches the behavior of the option added in node v6.3.0: https://nodejs.org/api/cli.html#cli_preserve_symlinks. See https://github.com/nodejs/node/issues/3402 for discussion around this feature in node.


devongovett:
@substack @feross who is maintaining browserify these days? Can someone look at this?

goto-bus-stop:
makes sense, thanks!

reading the node thread it seems like entry points _would_ be `realpath()`ed regardless, does this patch need to do anything handle that?

PR #1743, opened 2017-07-02
Add Node.js v8 to Travis CI

ryysud:
Add support for Node.js v8.
Updated .travis.yml.


abouthiroppy:
I think we can drop 0.12 and 0.10.

ryysud:
@abouthiroppy I think so too.

ryysud:
Please follow up on this 🙇

abouthiroppy:
@substack @maxogden What do you think?

.travis.yml
```diff
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
+  - "8"
   - "7"
```
stevemao: could you remove `7` and add `9`?

stevemao:
Note this will be a breaking change

ryysud:
@stevemao I fix it.

goto-bus-stop:
@ryysud sorry for the wait, curious if you could check out the test failure? it looks like it's a bug in the traceur version used by the `yield.js` test, maybe bumping the `es6ify` dependency in our package.json could work?

goto-bus-stop:
i'll remove es6ify after merging. thanks for the patch!

ljharb:
fwiw I think dropping node < 4 was a very very bad move, and I think that should be reverted.

ryysud:
🙇

ryysud:
@ljharb Sorry. Thank you for your Pull Request 🙇

PR #1755, opened 2017-08-27
Paraquire - paranoidal require

nickkolok:
Greetings from Russia, comrads!

I've prepared pull request using [paraquire](https://github.com/nickkolok/paraquire) to limit permissions of some dependencies of `browserify`. It will protect against some cases of dependency hijacking.
Unfortunately, in some cases it is unclear what right does the library need.

Please feel free to discuss and ask questions.


nickkolok:
Update: I see the problems with Node <= 0.12 and I'm working on compatibility or fallback

PR #1769, opened 2017-10-21
 Array support for b.exclude() and b.ignore()

goto-bus-stop:
Rebased @bakape's PR #1218 and updated it to use `isArray` instead of `isarray`, which is a thing that changed since the PR was opened 2 years ago.

Closes #1218


goto-bus-stop:
this got reasonable feedback previously so i'll go ahead and merge.

Issue #1782, opened 2017-12-15
Packages/paths in backticks (`) are ignored


TehShrike:
Browserify resolves `tape` correctly here:

```js
const test = require('tape')
```

but not here:

```js
const test = require(`tape`)
```

This differs from node's behavior.

---

Are you open to having this fixed?  I'm willing to add tests and write the patch.

## Context

I started using an eslint autoformat script with `quotes` set to `backtick`, so all of my strings are backtick-enclosed nowadays.  (See also: https://ponyfoo.com/articles/template-literals-strictly-better-strings )

goto-bus-stop:
Yeah, I think this would make sense!

TehShrike:
Once https://github.com/browserify/detective/pull/74 is merged and released, I'll update https://github.com/browserify/module-deps/pull/136 with the version bump, and then once that is merged and released, I'll update https://github.com/browserify/browserify/pull/1785

TehShrike:
Sweet!  Should I update the changelog so that a new release can go out?

PR #1785, opened 2017-12-20
Failing test + fix for backtick quotes

TehShrike:
Issue #1782

This issue needs to be fixed in `module-deps`.  I will open a pull request there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "inherits": "~2.0.1",
     "insert-module-globals": "^7.0.0",
     "labeled-stream-splicer": "^2.0.0",
-    "module-deps": "^4.0.8",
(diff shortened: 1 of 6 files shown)


goto-bus-stop:
in https://github.com/browserify/detective even, i think. thanks for working on this :D

TehShrike:
I'll update this to use `module-deps@2.x` and then it should be good.

TehShrike:
Bumping the dependency on `module-dep`s guarantees that the backtick-supporting version of `detective` will be used.

The tests fail in 0.10 and 0.12 because `detective` doesn't support those versions any more.

goto-bus-stop:
merging since the failures are only on node <4 and tests will pass on master. thanks!

PR #1793, opened 2018-01-04
Set the node version in package.json

TehShrike:
Testing against 0.10 and 0.12 was dropped in e8dbe6c8a7ffd88863ec8db3d6bfa51976be8c38

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,9 @@
     "type": "git",
     "url": "http://github.com/browserify/browserify.git"
   },
+  "engines": {
+    "node": ">= 4.0"
+  },
   "keywords": [
     "browser",
     "require",


TehShrike:
I'm down. 👍 

I know that whether or not dropping support for insecure versions of node should be a breaking change is a contentious topic, do you feel this should be a breaking or feature version bump?

Personally I tend towards breaking releases if dropping support for versions of node that are in maintenance, and feature/patch for dropping support for versions of node that are out of LTS, but I don't have strong opinions.

goto-bus-stop:
i think we should follow module-deps & detective and do a major bump :)

bcomnes:
Just make sure to update module-deps before doing the bump!

TehShrike:
The module-deps dependency got bumped in #1785, so it should be good to go.

Do y'all prefer having pull requests opened for changelog additions, or is it fine if I push changelog+package.json straight into master when publishing?

goto-bus-stop:
just pushing straight to master is how i've done it, so i'm cool w/ that!

PR #1799, opened 2018-01-15
Remove package-lock.json

goto-bus-stop:
Think this was added on accident at some point.


goto-bus-stop:
Oh cool! I didn't know npm supported project specific `.npmrc`s

PR #1801, opened 2018-01-31
Always resolve initial link

goto-bus-stop:
Follow up to #1742. Makes sure the entry point is always resolved, like in node.


alexeagle:
Node has the `--preserve-symlinks-main` option now. Should this change also be updated to conditionally allow that behavior? (Under Bazel, running browserify is getting wrong paths due to this change)

PR #1810, opened 2018-02-22
test/bare.js: fix file path

al-k21:
Fixes: https://github.com/browserify/browserify/issues/1809, https://github.com/nodejs/citgm/issues/545

### DIFF
--- a/test/bare.js
+++ b/test/bare.js
@@ -106,7 +106,7 @@ test('bare inserts __filename,__dirname but not process,global,Buffer', function
 test('bare inserts dynamic __filename,__dirname', function (t) {
     t.plan(2);
     
-    var file = 'dirname-filename.js';
+    var file = path.join(tmpdir, 'dirname-filename.js');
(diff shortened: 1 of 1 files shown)


al-k21:
I believe that my fix has nothing to do with the failing checks, please review @goto-bus-stop.

goto-bus-stop:
Yes the failures are unrelated. The tmpdir is necessary so that the file that the test writes gets cleaned up afterward.

I think the right fix here would be to pass `__dirname: process.cwd()` instead of `__dirname: path.dirname(file)` on line 126. That way we test if the paths are rebased correctly by checking if they match up with the original path.

al-k21:
> The tmpdir is necessary so that the file that the test writes gets cleaned up afterward.

Thanks, that makes sense. 
PR updated.

al-k21:
Hello, is this likely to get merged and released? @goto-bus-stop

PR #1815, opened 2018-03-06
ignore perf_hooks

yoshuawuyts:
Stub out the `perf_hooks` API so it doesn't crash when used with the `--node` flag. Thanks!


yoshuawuyts:
`v16.1.1`

PR #1822, opened 2018-03-30
Update coffeescript for tests

goto-bus-stop:
Avoids deprecation warnings.


goto-bus-stop:
Eeek, Node 6+ … maybe rather keep the deprecation warnings then for now. Could skip those tests but the warnings don't hurt so bad and I think coffee_bin is the only one that tests the `-c` option.

Issue #1746, opened 2017-07-13
When "main" matches source and "browser" field set, bundle fails.


benwiley4000:
Using browserify 14.4.0. masOS Sierra.

Quite a weird repro. When running this command:
```bash
browserify module.js -o dist/module.js
```
And under this set of circumstances:
* The `"main"` field in package.json matches the source file (`"module.js"`)
* The `"browser"` field is also set (with any value, but in our case `"dist/module.js"`)

We get this error when bundling:
```bash
Error: module not found: "/<pathroot>/module/module.js" from file /<pathroot>/module/_fake.js
```

Changing the `"main"` field to something else allows browserify to succeed. Removing the `"browser"` field also works. Of course these aren't options for distribution.

benwiley4000:
After some simple A/B testing it appears to me this issue was introduced by [v14.1.0](https://github.com/substack/node-browserify/compare/14.0.0...14.1.0).

benwiley4000:
The error above is emitted by the readonly stream returned by `b.bundle()`.

If I change [this line](https://github.com/substack/node-browserify/blob/master/bin/cmd.js#L64):

```javascript
bundle.pipe(fs.createWriteStream(tmpfile));
```
to:
```javascript
bundle.pipe(fs.createWriteStream(outfile));
```
and comment out this [this line](https://github.com/substack/node-browserify/blob/master/bin/cmd.js#L57):
```javascript
// bundle.on('end', successExit);
```
I have no issues.

Maybe at some stage the pipeline is expecting the stream to be written to a file matching the specified `outfile`?

I've had trouble finding why the `"main"` and `"browser"` fields in package.json would affect this.

benwiley4000:
I'd love to see some additional heads looking at this issue. It's quite a serious problem for browser-targeted distribution of a module written using ES2015+ syntax.

Since the syntax is fully supported in recent Node versions I don't want to point `main` to a transpiled copy, yet I also need a `browser` entry for the web.

ljharb:
You should always point "main" to a CJS ES5 copy; that's not going to change.

benwiley4000:
Sorry, I actually wasn't precise about (or actually stated wrongly) the issue: I want to be able to include a bundled copy of my library as the `browser` entry (separate from the `main` entry). For some reason browserify is no longer accommodating for the case where the `main` entry of an npm package is the browserify source and the `browser` entry is the browserify target.

[...]

PR #1834, opened 2018-05-15
Failing test for #1746

goto-bus-stop:
Not sure yet what causes this. The `basedir` option has to be set to the directory containing the package.json (or a child directory maybe, should also add a test for that).

It's trying to resolve the `bundle.js` file specified in the `browser` field, even though I tell it to use the full path to `main.js`. I might expect it to only use the `browser` field instead of the `main` field if I use the _directory_ as the entry point rather than the full file path.

### DIFF
--- a/test/browser_field_basedir.js
+++ b/test/browser_field_basedir.js
@@ -1,0 +1,16 @@
+var test = require('tap').test;
+var vm = require('vm');
+var browserify = require('../');
+
+test('browser field in basedir', function (t) {
+    t.plan(2);
+    var b = browserify(__dirname + '/browser_field_basedir/main', {
+        basedir: __dirname + '/browser_field_basedir'
+    });
+    b.bundle(function (err, src) {
+        if (err) { console.error(err); return t.fail() }
+        t.ifError(err);
+        vm.runInNewContext(src, { console: { log: log } });
+        function log (msg) { t.equal(msg, 'ok') }
+    });
+});
--- a/test/browser_field_basedir/main.js
+++ b/test/browser_field_basedir/main.js
@@ -1,0 +1,1 @@
+console.log('ok')
--- a/test/browser_field_basedir/package.json
+++ b/test/browser_field_basedir/package.json
@@ -1,0 +1,4 @@
+{
+    "browser": "bundle.js",
+    "main": "main.js"
+}


smoke:
I confirm this test replicates the issue https://github.com/browserify/browserify/issues/1746 and I am currently looking for a solution.

smoke:
@goto-bus-stop I have invested some time trying to fix this issue, however things are quite complicated and I couldn't figure out any reasonable solution.

I think there are couple of options to fix that:
 - as you have mentioned here https://github.com/browserify/browserify/issues/1746#issuecomment-389150933 ignore the `browser` field in case the examined code is out of `node_modules` - looks ok as approach - will probably work for most of the packages, yet it might be a breaking for repositories containing many packages in one and the same repository, relying on this to work in the way it does.
 - other approach I can think of is to have some option to to specify exact entries, that skip the package.json resolving
 - I have also tried working around the issue using `packageFilter` (see below), but this function was not used from `browser-resolve`, it was used only from `resolve`:
```javascript
var test = require('tap').test;
var vm = require('vm');
var browserify = require('../');

test('browser field in basedir', function (t) {
    t.plan(2);
    var b = browserify(__dirname + '/browser_field_basedir/main', {
        basedir: __dirname + '/browser_field_basedir',
        packageFilter: function packageFilter(p, dir) {
            if (!dir.includes('node_modules')) {
                const packageInfo = {
                    ...p
                };
                delete packageInfo['browser'];
                return packageInfo;
            }
            return p;
        }
    });
    b.bundle(function (err, src) {
        if (err) { console.error(err); return t.fail() }
        t.ifError(err);
        vm.runInNewContext(src, { console: { log: log } });
        function log (msg) { t.equal(msg, 'ok') }
    });
});

```

Sorry but, so far I am giving up on this effort and will probably try local for the project workaround, or try switching https://github.com/dtjohnson/xlsx-populate build to use webpack instead of browserify so that builds run with defined `browser` field in `package.json`

goto-bus-stop:
Understandable; thanks for the research! I'll look into this more this week

goto-bus-stop:
This is pretty tough; i think sometimes it _is_ expected that browserify takes the `"browser"` field locally. I'm not sure how to reconcile that. technically `"browser"` isn't really meant to be used for built bundles.

ljharb:
Definitely not; "browser" is just like "main for browsers" - an alternative *entry point*. Entry points are not bundles; they are *bundled*.

PR #1841, opened 2018-06-09
Add Awesome Mentioned Badge

bsoyka:
This repository has been mentioned in the Awesome JavaScript list, and it deserves a badge.


goto-bus-stop:
I appreciate the PR (and the praise!) but I don't think that a badge like that makes much sense for browserify's technical readme.

PR #1844, opened 2018-06-15
Upgrade util to v0.12.0.

goto-bus-stop:
This release includes `util.promisify` and `util.callbackify` support. Users have to provide their own Promise polyfill if they use those functions while their target environment doesn't support it.
No breaking changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "timers-browserify": "^1.0.1",
     "tty-browserify": "0.0.1",
     "url": "~0.11.0",
-    "util": "~0.10.1",
+    "util": "~0.12.0",
     "vm-browserify": "^1.0.0",
     "xtend": "^4.0.0"
   },
--- a/test/util.js
+++ b/test/util.js
@@ -1,6 +1,7 @@
 var browserify = require('../');
 var test = require('tap').test;
 var util = require('util');
+var xtend = require('xtend');
 var vm = require('vm');
 
 test('util.inspect', function (t) {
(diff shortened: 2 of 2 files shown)


mattdesl:
Seems good! Any chance of merging & cutting a new release for this?

ljharb:
using https://npmjs.com/util.promisify would handle it for the cases where they were using older node, or browsers, fwiw

goto-bus-stop:
@ljharb I checked that out initially but since `util` already requires ES5 for other reasons, I didn't think it would be worth the extra `es-abstract` kBs for most users. When ES3 support is required users already have to be very careful about the (core) modules they use, and they can trivially install the `util.promisify` module separately.

I'd like to merge and release this soon, if you don't strongly object to the above :)

ljharb:
That’s fine, but we could alias out the es-abstract parts if needed.

shinnn:
@goto-bus-stop Can you update the PR title to "Upgrade util to v0.12.0."? I was about to point out that util v0.12 has been released, but in fact you've already updated this PR accordingly, too. 27d74e08ecf558cb5ba89a6823b88587128651e4

RangerMauve:
This'd be super nice to have since the `promisify` method isn't currently in browserify

RangerMauve:
Is there anything I could do to help? Like opening a new PR

Raynos:
Can we cut and publish a new version of `browserify` that includes util @ 0.12.0 ?

PR #1865, opened 2018-08-31
Use the latest punycode(1.3.2) 

oldmtn:
Use the latest punycode, since punycode(1.3.2) export with:
`define('punycode', function() {
    return punycode;
});
`
which isn't portable when using requireJS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "parents": "^1.0.1",
     "path-browserify": "~0.0.0",
     "process": "~0.11.0",
-    "punycode": "^1.3.2",
+    "punycode": "^2.1.1",
     "querystring-es3": "~0.2.0",
     "read-only-stream": "^2.0.0",
     "readable-stream": "^2.0.2",


oldmtn:
> punycode 2.0.0 is actually identical to 1.3.2 except for the ES6 syntax.

I don't think so.
// punycode, 1.3.2
`define('punycode', function() {
    return punycode;
});`

// punycode, 2.0.0
`/** Define the public API */
const punycode = {
	/**
	 * A string representing the current Punycode.js version number.
	 * @memberOf punycode
	 * @type String
	 */
	'version': '2.0.0',
	/**
	 * An object of methods to convert from JavaScript's internal character
	 * representation (UCS-2) to Unicode code points, and back.
	 * @see <https://mathiasbynens.be/notes/javascript-encoding>
	 * @memberOf punycode
	 * @type Object
	 */
	'ucs2': {
		'decode': ucs2decode,
		'encode': ucs2encode
	},
	'decode': decode,
	'encode': encode,
	'toASCII': toASCII,
	'toUnicode': toUnicode
};

module.exports = punycode;`

Yes, browserify is mainly target convert ES6 to ES5, if browserify itself contain ES6 code.
Than its verfy difficult run in old browser.

Since you can't change it here. Let me try talk to punycode(1.3.2) to remove the uncomfortable amd id('punycode').

oldmtn:
Besides, 
`The best thing to do here might be to have a new punycode-browserify package that is punycode 2.0.0 compiled to ES5.`

Could you help to do it?

goto-bus-stop:
Yes I think I worked on one a few months ago. I'll see if I can dig it up on the weekend.

ljharb:
I'm confused what the problem is with the AMD detection; couldn't that be trivially wrapped for CJS?

Issue #1873, opened 2018-10-26
Using --list with --require gives unexpected results


casr:
As per the documentation, `--list` seems to be a good option for use with Makefiles. I have noticed that with `--require`, however, the module name is used instead of its file path. This means that for a Makefile the dependencies are not properly established. For example:

```sh
# setup
$ mkdir -p /tmp/test ; cd /tmp/test ; npm i browserify

$ browserify --list --require path
/tmp/test/node_modules/process/browser.js
path
```

Notice how "path" appears instead of the full file path compared with:

```sh
$ echo "require('path')" | browserify --list -
/tmp/test/node_modules/process/browser.js
/tmp/test/node_modules/path-browserify/index.js
/tmp/test/_stream_0.js
```

Seems like a bug?

PR #1874, opened 2018-10-26
Provide test cases for --list

casr:
Before attempting to fix this I thought I would provide a test for the current behaviour and one for the failing behaviour from #1873 

It would be good to understand if this is the right approach to testing this sort of thing.

Before merging it might be nice to fix the feature but if this takes longer than expected then one test can go in and the other be skipped until someone has the time to fix it


goto-bus-stop:
Super, thanks :+1: the test looks good to me from just eyeballing it, not sure off the top of my head if the order is guaranteed. I think the solution to this problem will involve setting the `row.file` property correctly for `--require/b.require()` modules either here or in module-deps.

casr:
Thank you for having a look and a good note about ordering. I guess I could sort before comparing in the test unless the ordering is guaranteed elsewhere. I will add a note if I find that to be the case.

PR #1876, opened 2018-10-31
#1875 - update package with buffer@5.2.1

### LINKED ISSUE #1875: Updating to latest Buffer polyfill
At this time, browserify@16.2.3 relies on outdated version of buffer polyfill 
    "buffer": "^5.0.2",
Some bugs have been fixed in buffer@5.2.1 ( ArrayBuffer.isView error)

erossignon:
Fixes #1875

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "browser-pack": "^6.0.1",
     "browser-resolve": "^1.11.0",
     "browserify-zlib": "~0.2.0",
-    "buffer": "^5.0.2",
+    "buffer": "^5.2.1",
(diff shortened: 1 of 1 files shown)


erossignon:
semver : I thought so as well however after installing browserify on a fresh install it looks like buffer is still stuck to version  5.0.1 => may be automatic version doesn't apply to minor version but only patch number.

ljharb:
The carat applies to minors as well. Make sure you don’t have a lockfile.

erossignon:
@ljharb  you're right "^5.0.1" resolved to "5.2.1"

Issue #1539, opened 2016-04-01
Package replacement as a browser field


dfcreative:
Hello @substack @feross @zertosh @hughsk!

There is a chance I’m missing something, but is there any specific reason why we cannot implement package replacement in `browser` field, not only the file replacement?

```json
"browser": "package-analog-for-browser",
"browser": {
  "some-node-package": "some-browser-package"
}
```
instead of 

```json
"browser": "./package-shim.js",
"browser": {
  "some-node-package": "./some-node-package-shim.js"
}
```

Often I see a situation where packages are just being targeted to single platform and have 1:1 analog in other platform, e. g: 
* [webaudio](https://www.npmjs.com/package/webaudio) / [baudio](https://www.npmjs.com/package/baudio)
* [gl-shader-output](https://www.npmjs.com/package/gl-shader-output) / [nogl-shader-output](https://www.npmjs.com/package/nogl-shader-output)
* [audio-speaker](https://www.npmjs.com/package/audio-speaker) / [node-speaker](https://www.npmjs.com/package/node-speaker)
* [glslify](https://www.npmjs.com/package/glslify) / [glslify-sync](https://www.npmjs.com/package/glslify-sync) for node.
* any of the [browser-builtins](https://github.com/substack/node-browserify/blob/master/lib/builtins.js) etc.

In some cases, like the [glslify-sync](https://github.com/vorg/glslify-sync), absence of that feature is really painful, for if we had the opportunity to set package replacement, we could avoid [duplicating code](https://github.com/vorg/glslify-sync/blob/master/transform.js) and keep consistent API.

There are modules like aliasify, but they are... too opinionated and do not actually solve the cases like above.

feross:
That is already supported. Here's an example of it in use: https://github.com/feross/webtorrent/blob/84052ad40dc6b788bfc173453a67da155e7804bc/package.json#L14

dfcreative:
@feross it does not recognize module as the main browser pointer:

```js
"browser": "targtet-package"
```

dfcreative:
@feross if it is unintended behaviour, please reopen to verify that its a bug and let me make a PR.
Or should I address it in separate issue?

feross:
Yeah, not sure if this is intentional or not.

dy:
Ok, the issue will be fixed with https://github.com/defunctzombie/node-browser-resolve/pull/84

PR #1887, opened 2018-11-28
Browser resolve direct

dy:
Adds #1539 test for now.
Will be fixed once https://github.com/defunctzombie/node-browser-resolve/pull/84 is merged.


index.js
```diff
@@ -40,7 +40,7 @@ function Browserify (files, opts) {
     var self = this;
     if (!(this instanceof Browserify)) return new Browserify(files, opts);
     if (!opts) opts = {};
-
```
ljharb: please revert all these unrelated whitespace changes.

dy:
Sublime automatically cleans up trailing whitespaces. One moment.

index.js
dy: Sublime automatically cleans up trailing whitespaces. Not sure I understand what is the meaning of keeping them.
Bah, sad to see there is no way for my contributions.

goto-bus-stop: i wouldn't keep them in most situations, but browserify is old and most of the code was written by people who no longer maintain it, so history tools like `git blame` are important for people like me who have no idea why things are the way they are a lot of the time :sweat_smile: minimal patches are valuable in that respect.

thanks for the test, i hope we can get the browser-resolve patch in soon.

PR #1913, opened 2019-07-05
Add http2 to builtins

steve-taylor:
Node.js 10.10.0 added the `http2` module. This change adds just enough support for `--no-builtins` (and therefore `--node`) to recognise `http2` as a Node.js built-in module and not attempt to bundle it.

### DIFF
--- a/lib/builtins.js
+++ b/lib/builtins.js
@@ -12,6 +12,7 @@ exports.events = require.resolve('events/');
 exports.fs = require.resolve('./_empty.js');
 exports.http = require.resolve('stream-http');
 exports.https = require.resolve('https-browserify');
+exports.http2 = require.resolve('./_empty.js');
 exports.inspector = require.resolve('./_empty.js');
 exports.module = require.resolve('./_empty.js');
 exports.net = require.resolve('./_empty.js');


steve-taylor:
You're welcome. Thanks for the quick merge. For now, I'm adding `-u http2` to the command line.

ljharb:
It might be worth using resolve’s listing of core modules to default all of them to empty (and override with shims as needed) - then this would have Just Worked.

goto-bus-stop:
that sounds like a good idea—we've had a few browserify releases lately that were mostly about adding new empty shims.

ljharb:
(Pull in `resolve/lib/core`, enumerate the entries, and any key whose value is true is a core module in the current node version)

steve-taylor:
The set of core modules isn’t the same for all Node versions. We might not necessarily build and run on the same version of Node.

Maybe we need an extra option to specify the target runtime, e.g. `—node-version 10.16.0`.

ljharb:
@steve-taylor that’s a reasonable addition to resolve if it’s needed; but i do think that building and running on the same version is typical.

PR #1916, opened 2019-08-07
Upgrade stream-http to v3

feross:
This version of stream-http drops support for IE10 and below.

This should not require a major version bump since browserify 14.0.0 already upgraded the buffer package and dropped support for IE8-10. Since stream uses buffer internally, stream is already not working in IE8-10.

This lets us get the latest and best improvements.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "shasum": "^1.0.0",
     "shell-quote": "^1.6.1",
     "stream-browserify": "^2.0.0",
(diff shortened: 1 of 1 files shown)


feross:
The build is passing even though GitHub isn't showing it: https://travis-ci.org/browserify/browserify/builds/568633234

Going to merge this and release a new version.

feross:
16.4.0

PR #1918, opened 2019-08-08
Support custom names for "browser" field resolution

feross:
Fixes: https://github.com/browserify/browserify/issues/1917

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,12 +83,19 @@ function Browserify (files, opts) {
     self._transforms = [];
     self._entryOrder = 0;
     self._ticked = false;
-    self._bresolve = opts.browserField === false
+
+    var browserField = opts.browserField
+    self._bresolve = browserField === false
         ? function (id, opts, cb) {
             if (!opts.basedir) opts.basedir = path.dirname(opts.filename)
             resolve(id, opts, cb)
         }
-        : bresolve
+        : typeof browserField === 'string'
+            ? function (id, opts, cb) {
+                opts.browser = browserField
+                bresolve(id, opts, cb)
+            }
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -83,12 +83,19 @@ function Browserify (files, opts) {
...
-    self._bresolve = opts.browserField === false
+
+    var browserField = opts.browserField
+    self._bresolve = browserField === false
         ? function (id, opts, cb) {
             if (!opts.basedir) opts.basedir = path.dirname(opts.filename)
             resolve(id, opts, cb)
         }
-        : bresolve
+        : typeof browserField === 'string'
```
ljharb: are we sure nobody is putting a string value now, and relying on it being a noop?

What if it's an empty string?

feross: Right now any value other than `false` (including an empty string) just does normal `"browser"` field resolution.

We've never supported any value other than `false` for this argument and since this is an additive change, I'd argue it could released in a minor version. But I'm also fine with releasing it in a major version to be extra safe, though I think it's a bit overkill.

ljharb: Sorry, that was two questions :-) you answered the first, but not the second - after this PR, should an empty string be allowed?

feross: The simplest rule would be: if the type is `string`, then we look for that key in `package.json` and apply the "browser" field spec.

So, then an empty string would look for an empty string key in `package.json`.

But it's also likely to be a bug, so also I'm okay with throwing an exception in this case.

feross:
@substack took a look at this and said it looks good on IRC. I'm going to merge it now.

feross:
16.5.0

PR #1932, opened 2019-11-11
Polyfill Object.setPrototypeOf in tests on old Node.js versions

goto-bus-stop:
The `buffer` shim needs `setPrototypeOf` since https://github.com/feross/buffer/pull/238. `buffer`'s official support targets had it. Node 0.8/0.10 incidentally had the other things it needs, but not this.

### DIFF
--- a/test/buffer.js
+++ b/test/buffer.js
@@ -3,6 +3,7 @@ var test = require('tap').test;
 var vm = require('vm');
 
 if (!ArrayBuffer.isView) ArrayBuffer.isView = function () { return false; };
+if (!Object.setPrototypeOf) Object.setPrototypeOf = require('setprototypeof');
 
 test('utf8 buffer to base64', function (t) {
     t.plan(1);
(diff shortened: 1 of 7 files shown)


test/buffer.js
```diff
@@ -3,6 +3,7 @@ var test = require('tap').test;
 var vm = require('vm');
 
 if (!ArrayBuffer.isView) ArrayBuffer.isView = function () { return false; };
+if (!Object.setPrototypeOf) Object.setPrototypeOf = require('setprototypeof');
```
ljharb: instead of polyfilling it, should it be imported in the implementation? otherwise tests will be assuming the user has polyfilled it, which is not guaranteed.

goto-bus-stop: I think we can import it in the files that are being bundled for these tests, yes. `buffer` itself probably wouldn't want to add it because all of the browsers that they support have `setPrototypeOf`.

goto-bus-stop:
addressed by https://github.com/browserify/browserify/pull/1939

PR #1937, opened 2019-12-27
gitignore - add node_modules

kumavis:
the original author removed `node_modules` from the `.gitignore` a few times in the past, I think its generally considered a good idea now :+1:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 package-lock.json
+node_modules


kumavis:
test failures are unrelated

kumavis:
seems like `Object.setPrototypeOf` is used in some dependency, which is not available in node v0.10 and v0.8

ljharb:
Sounds like we need to peg that dep and file upstream issues, before landing additional PRs.

kumavis:
lol it was me https://github.com/feross/buffer/commit/db5572c747f3f73ea84ac85729c7527db7016ec2

kumavis:
fix here https://github.com/browserify/browserify/pull/1939

kumavis:
@ljharb ready

PR #1939, opened 2019-12-28
Fix support for node v0.8 support

kumavis:
@ljharb


package.json
```diff
@@ -28,7 +28,7 @@
     "browser-pack": "^6.0.1",
     "browser-resolve": "^1.11.0",
     "browserify-zlib": "~0.2.0",
-    "buffer": "^5.0.2",
+    "buffer": "5.2.1",
```
ljharb:
```suggestion
    "buffer": "~5.2.1",
```

package.json
```diff
@@ -84,7 +84,7 @@
     "semver": "^5.5.0",
     "seq": "0.3.5",
     "tap": "^10.7.2",
-    "temp": "^0.8.1",
+    "temp": "0.8.3",
```
ljharb:
```suggestion
    "temp": "~0.8.3",
```
?

kumavis: the breakage here is in `0.8.4`

ljharb: edited to show the intent with = then :-)

kumavis:
I also noticed appveyor and travis tests dont have the same engine coverage...

kumavis:
would be nice to document supported platforms for both browserify and bundles

kumavis:
also if we could turn on greenkeeper for only in-range dep updates? would give us an idea if our deps break

kumavis:
woof appveyor takes forever

kumavis:
@ljharb finished cooking

ljharb:
We should definitely make sure “engines” is set correctly, as are CI matrices.

PR #1988, opened 2020-11-24
BREAKING: Update buffer to ^6.0.3

feross:
Update buffer to ^6.0.3

## Browser support

BREAKING: Drop IE11, Safari 9-10 support. This affects ~1.16% of global browser users (source: https://caniuse.com/usage-table) and this number continues to decrease each month.

Since `buffer` now supports the newly added `BigInt` methods from Node.js and the implementation uses the exponentiation operator throughout, we decided to drop support for IE11 and Safari 9-10 from `buffer`.

Similar to previous `buffer` upgrades (e.g. https://github.com/browserify/browserify/pull/1678), users can manually specify an older `buffer` version if they need IE11 and Safari 9-10 support. More information about how to do this can be found in: https://github.com/browserify/browserify/issues/1980

## Semver

This should be released as semver major.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "browser-pack": "^6.0.1",
     "browser-resolve": "^2.0.0",
     "browserify-zlib": "~0.2.0",
-    "buffer": "~5.2.1",
+    "buffer": "^6.0.3",
     "cached-path-relative": "^1.0.0",
     "concat-stream": "^1.6.0",
     "console-browserify": "^1.1.0",


andreibancioiu:
For those that need to reference this new version of `buffer` within their `browserify` pipeline (until the PR is merged etc.), this might help:

```
# Install latest version of buffer
npm install buffer --save-dev 

# Use --require to bundle using the latest version of buffer
browserify foobar/*.js --require buffer/:buffer -o foobar-bundle.js --standalone foobar
```

goto-bus-stop:
Nice. I think dropping IE 11 is quite a significant step (moreso than any other IE version), so I'd prefer to make it easier to swap out builtins consistently in both the API and CLI before we actually publish this.

feross:
@goto-bus-stop I think we dropped IE8-10 when it had a similar amount of usage (see: https://github.com/browserify/browserify/pull/1678) but if it makes you more comfortable, I'm okay with waiting until usage has declined further or it has become easier to swap out built-ins in the API. :)

Btw, since I sent this PR, IE11 usage has declined from 1.16% to 1.09%. https://caniuse.com/usage-table

ljharb:
1% is still 4.6 million people - I'd prefer we don't reduce human beings to percentages.

feross:
Just a quick note that since I sent this PR, IE11 usage has declined from 1.16% to 1.00%.

> @ljharb 1% is still 4.6 million people - I'd prefer we don't reduce human beings to percentages.

Lol, it goes without saying that each human being is unique and special. By that logic, we can't ever drop support for anything because we can't "reduce human beings to percentages". We've dropped support for browsers when they reached similar usage numbers in the past. See: https://github.com/browserify/browserify/pull/1678 Also, keep in mind that releasing a new major version of `browserify` doesn't mean all these users are suddenly abandoned. Developers can decide whether or when to upgrade `browserify` and drop IE11 support. The current version of `browserify` will continue working just fine.

ljharb:
Unless future features and bugfixes are backported to that version of browserify, it does indeed mean they miss out on those.

goto-bus-stop:
I don't think I'll have the time and interest to work on a feature for replacing builtins in plugins, so maybe the next best thing is to merge this as is and document that users can do `npm install buffer@5` and `--require buffer/:buffer` to maintain support for IE 11.

jimmywarting:
0.85% now 😄

DiegoRBaquero:
0.76%, we should set the threshold at 0.5%

Issue #1999, opened 2021-03-18
multiple stream versions galore


pkit:
Is there any reason why `stream-browserify` depends on `readable-stream@3.6.0` while `browserify` itself depends on `readable-stream@^2` ?
Right now streams are kind of unusable:
1. `stream-browserify` dependency tree is botched see here https://github.com/browserify/stream-browserify/issues/25
2. Using `readable-stream` directly leads to multiple versions in the bundle which in turn leads again to no inheritance anywhere.
3. `readable-stream` has `inspect` bug here https://github.com/nodejs/readable-stream/pull/435#issuecomment-801867019 which prevents me from debugging **why** there is any difference in `Stream` prototypes...

If there is no particular reason for such behavior I can create PR(s) that fix it, but maybe I'm missing something?

PR #2070, opened 2023-09-10
Update readable-stream to v3

legobeat:
This is a revival of a past attempt (#1938) which was partially applied in #1970 .

This follows up to also update the direct dependency on `readable-stream` to v3. 

#### Resolves

- #1880
- #1999


ljharb (changes requested):
An upgrade that forces us to drop support of a node version is an upgrade not worth doing.

PR #2008, opened 2021-05-09
update Node.js versions used in Travis CI checks

striezel:
Node.js versions below 10.x are not maintained anymore, so it is probably best to use some newer, still maintained versions.
For release cyles of Node.js see the official site: <https://nodejs.org/en/about/releases/>.

PR #2071, opened 2023-09-10
Change minimum node.js version from 0.8 to 4

legobeat:
Motivation: `readable-stream@3` [breaks compatibility](https://ci.appveyor.com/project/browserify/browserify/builds/47996942) with Node.js versions < 4.

The last version to reach EoL being dropped here is v0.12.x, which ended maintenance 2016-12-31 .

#### Blocking:
- #1880 

#### Related:
- #2008

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "url": "http://github.com/browserify/browserify.git"
   },
   "engines": {
-    "node": ">= 4"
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
I see no reason to drop anything; if readable-stream v3 doesn’t support the node versions we do, we can simply not upgrade to it.

legobeat:
@goto-bus-stop WDYT?

PR #2094, opened 2026-07-19
Refactor: use globv7 instead of glob

tj-commits:
Glob v7 is no longer maintained, and [I created a time back a new fork of Glob v7](https://github.com/legacy-semvers/globv7) that will be maintained in case a vulnerability is discovered.

Glob v7 is depended on by [fresh-tape](https://github.com/martinheidegger/fresh-tape), a library that is depended on for testing by [testem](https://github.com/testem/testem), a package with 200,000 weekly downloads.

This PR refactors to use my maintained fork instead of the old one.
As quoted by isaacs, "Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me".


ljharb:
I'm going to decline this; the issues with glob v7 don't affect us since browserify is not a long-running process, and depending on a version with [131M weekly downloads](https://majors.nullvoxpopuli.com/q?packages=glob) with a minor problem is a far more responsible and secure choice than depending on a package with less than a thousand.

(the same logic applies to fresh-tape or testem vs tape, for example)

PR #2097, opened 2026-09-04
ci: port from appveyor and travis

evenstensberg:


.github/actions/test/action.yml
```diff
@@ -0,0 +1,52 @@
...
+
+runs:
+  using: composite
+  steps:
+    # The npm shipped with the older node versions can no longer reach the
+    # registry (TLS), which is what travis worked around with
+    # `nvm install-latest-npm`. Install with a current npm instead, then
+    # switch to the version under test to actually run the suite.
+    - name: install dependencies (current node)
+      uses: actions/setup-node@v4
```
ljharb, line 19: this is fine, but i personally prefer using ljharb/actions/node/install, which uses nvm and ime works much better than setup-node. (it also uses nvm, and `nvm install-latest-npm`)

.github/actions/test/action.yml
```diff
@@ -0,0 +1,52 @@
...
+    # own cache to key off; cache the npm download dir on package.json instead.
+    - name: locate npm cache
+      id: npm-cache
+      shell: bash
+      run: echo "dir=$(npm config get cache)" >> "$GITHUB_OUTPUT"
+    - uses: actions/cache@v4
+      with:
+        path: ${{ steps.npm-cache.outputs.dir }}
+        key: ${{ runner.os }}-npm-${{ hashFiles('package.json') }}
+        restore-keys: ${{ runner.os }}-npm-
```
ljharb, line 33: is this actually necessary? unless the speed is prohibitive without it, we're better off avoiding the complexity (and cache poisoning attack vector) of caching

PR #2098, opened 2026-09-05
docs: refresh

evenstensberg:


bin/advanced.txt
```diff
@@ -4,15 +4,15 @@ Advanced Options:
 
     Skip detection and always insert definitions for process, global,
     __filename, and __dirname.
-                  
+
     benefit: faster builds
     cost: extra bytes
-
```
ljharb, line 10: probably worth having a separate single commit that purges all trailing whitespace in the entire repo?

readme.markdown
```diff
@@ -9,10 +9,42 @@ browserify will recursively analyze all the `require()` calls in your app in
...
 tag.
 
-[![build status](https://img.shields.io/travis/browserify/browserify/master.svg)](https://travis-ci.org/browserify/browserify)
+[![npm version](https://img.shields.io/npm/v/browserify.svg)](https://www.npmjs.com/package/browserify)
+[![downloads](https://img.shields.io/npm/dm/browserify.svg)](https://www.npmjs.com/package/browserify)
+[![license](https://img.shields.io/npm/l/browserify.svg)](./LICENSE)
 
 ![browserify!](./assets/logo.png)
 
+# table of contents
```
ljharb, line 18: is this necessary, given that github automatically generates a TOC on every markdown file?

readme.markdown
```diff
@@ -60,11 +92,21 @@ html!
...
-npm install browserify
+npm install --save-dev browserify
+```
+
+That gives you the `browserify` command in `node_modules/.bin`, so you can run
+it from an npm script or with `npx browserify`. To put the command on your
+`PATH` everywhere instead, install it globally:
+
+```
+npm install -g browserify
```
ljharb, line 103: let's not advice people to install anything globally; the modern advice is to never install anything globally, and to use `npx`.

readme.markdown
```diff
@@ -826,6 +925,39 @@ add a package.json keyword of `browserify-tool` so that
...
+of the same name holding the fixture files that test bundles. When you add a
+feature, add a test file next to the existing ones.
+
+Everyone participating in the project is expected to follow the
+[code of conduct](code-of-conduct.md).
+
+# security
+
+Please report security vulnerabilities as described in the
+[security policy](security.md) rather than in a public issue.
```
ljharb, line 959: this is fine, but seems redundant to me since https://github.com/browserify/browserify/security/policy already exists

PR #2099, opened 2026-09-05
chore: add nvmrc file

evenstensberg:


ljharb (approved):
not sure if we actually want people using this for active development, but it doesn't hurt

evenstensberg:
@ljharb looks like you need to remove travis from the settings page directly

ljharb:
should be all set

PR #2100, opened 2026-09-05
feat: ts support

evenstensberg:


package.json
```diff
@@ -85,7 +85,8 @@
     "seq": "0.3.5",
     "tap": "^10.7.2",
     "temp": "=0.8.3",
-    "through": "^2.3.4"
+    "through": "^2.3.4",
+    "typescript": "^5.0.0"
```
ljharb, line 89: why v5? v5 still requires node 12, so it's already violating engines.node - why not use v6 or v7?

